PR #93, opened 2025-01-21
Add support for URLs in import command

### LINKED ISSUE #80: Able to create a new job
### Reason/Context

To be able to create a new job import from specifications available on other servers

### Description

I am using a lot of specifications available on github enterprise servers. 
With the microcks UI I can add them with the _Importers -> Create a new Job_  feature but in the cli I can only import files that are located on the machine that executes the command.

### Implementation ideas

Maybe the microcks-cli import function can be extended to accept also URLs

Harsh4902:
### Description

- This PR adds support for URLs in `import` command.
- Now `microcks-cli` is able to fetch artifacts or specification file from the URL and import it into the Microcks.

**As of now, we can only import artifacts from the URL as a primary artifact.

### Changes

- Updated `UploadArtifact` function of `MicrocksClient`.
- Updated `ImportCommand` which will accept artifact from URL as a Primary Artifact.
- `ioutil.ReadAll()` is deprecated, which is replaced by `io.ReadAll`.

### Related issue(s)
Resolves #80

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,6 @@
 
 # Ouput of the build-binaries script
 build/_output
+
+# .DS_Store file of MacOS
+.DS_Store
--- a/cmd/help.go
+++ b/cmd/help.go
@@ -40,6 +40,7 @@ func (c *helpCommand) Execute() {
 	fmt.Println("  help        display this help message")
 	fmt.Println("  test        launch new test on Microcks server")
 	fmt.Println("  import      import API artifacts on Microcks server")
+	fmt.Println("  import-url  import API artifacts from URL on Microcks server")
 	fmt.Println("")
 	fmt.Println("Use: microcks-cli test <apiName:apiVersion> <testEndpoint> <runner> \\")
 	fmt.Println("   --microcksURL=<> --waitFor=5sec \\")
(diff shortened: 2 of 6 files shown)


Harsh4902:
@lbroudoux Please review this PR.

lbroudoux:
Hey @Harsh4902 

While being correct, you've not been the easiest path to take to do that. The Microcks API also has a `/api/artifact/download` endpoint that supports importing a remote http/https Artifact. This endpoint also have a `main` and a `secretName` attributes that allow managing the authentication while accessing the file on the remote server.

I think it would be more convenient to add a new command (like `import-url`) that will invoke a new `DownloadArtifact` function on Go Microcks client in order to do that.

Something like `microcks-cli import-url http://acme.com/openapi.yaml:true:my-secret` or `microcks-cli import-url http://acme.com/openapi.yaml` if it's a primary artifact that is publicly accessible. What do you think?

Harsh4902:
@lbroudoux Oh! I was unaware about `/api/artifact/download` endpoint. Thank you for suggesting this, I will make changes ASAP.

Harsh4902:
@lbroudoux Can we use a flag like `--url` inside import command to specify that a given artifact is a URL? I think it will be more convenient rather than creating a new command. What you say?

PR #109, opened 2025-02-16
Start and Stop command using cobra cli

Luxxgit2k4:
Adding the start and stop commands for microcks-cli using cobra cli. The modules have also been updated

microcks-cli start : Starts the server in the background by pulling the docker image with the container name "microcks-server"

microcks-cli stop: Stops the  server. 

Also updated main.go to prevent overwriting the existing commands. 

Future implementation:
Adding viper for managing configuration, for example allowing the users to select the port to run the microcks server.
Migrating the other commands into cobra cli for easy management. Cobra cli generates a help page based on the commands built without even need for building the help command.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -1,0 +1,28 @@
+package cmd
+
+import (
+  "fmt"
+  "os"
+  "github.com/spf13/cobra"
+)
+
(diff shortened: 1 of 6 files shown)


lbroudoux:
Now done in `1.x` branch.

PR #110, opened 2025-02-19
Enhance GoReleaser config with multi-platform builds, release customizations, and changelog filters

karanngi:
This PR updates the GoReleaser config to support:

1. Multi-platform builds (linux, darwin, windows) and multiple architectures (amd64, arm64).
2. Custom archive formats (tar.gz for most, zip for Windows).
3. A before hook to run go mod tidy before builds.
4. A release name template for better versioning.
5. Changelog filters to exclude commit types like docs, tests, and typos.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -1,13 +1,42 @@
 version: 2
+project_name: microcks-cli
+
+before: 
+  hooks:
+    - go mod tidy
 
 builds:
   - main: ./main.go
     binary: microcks-cli
     env:
       - CGO_ENABLED=0
+    goos: 
+      - linux
+      - darwin
+      - windows
+    goarch:
+      - amd64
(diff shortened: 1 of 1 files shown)


.goreleaser.yaml
```diff
@@ -1,13 +1,42 @@
...
+
+archives:
+- format: tar.gz
+  format_overrides:
+    - goos: windows
+      format: zip
 
 release:
+  name_template: "Microcks CLI v{{.Tag}}"
+  draft: true
```
lbroudoux, line 34: Why putting this as a draft? Do you know what is actually the default value? Does not look clear when looking at https://goreleaser.com/customization/release/#github

lbroudoux (commented):
Just got a question on draft mode for release but it looks all good to me.

PR #143, opened 2025-04-11
Feat: Add support for `start` and `stop` command

Harsh4902:
### Description
- This PR adds support for `start` and `stop` commands to CLI.
- Users can now start microcks instance using `microcks start` and can stop it using `microcks stop`.
- Also adds support for `config.yaml` file which will be used to store configs of microcks instance.
- It also supports `--driver` flag, to specify which driver will be used to start the container/microcks instance.

### Related issue(s)
Fixes #108

PR #157, opened 2025-05-12
Feat: Add support for config file and context switching

### LINKED ISSUE #136: Keep track of Microcks instance using something like `context file`
### Reason/Context

- We are going to add support for `start` and `stop` command for our CLI tool, which will help to start and stop a microcks container.
- For this feature, we have to keep track of existing microcks container so that en execution of `start` or `stop` command can be efficient.

### Description

- Execution of `start` or `stop` command will may lead us toward some failure.
- To prevent this, we have to keep track of existing microcks container using something like `context file`.

### Implementation ideas

_No response_

Harsh4902:
### Description

- This PR add support for `config` file which will store all the Microcks configurations and contexts related information.
- It also adds support for `context` command, which will enable users to switch between multiple context and delete a context.
- As of now, there is no way to create a context. A context will be created once the user will perform a login operation.
- This PR also adds `errors/error.go` package which will help to check errors in our CLI tool and handle them appropriately.

** We have added support for `config.yaml` file with #143. With this PR we are going to add a new `localconfig.go` which will manage all the `config` related things. `config.go` file will be removed in upcoming PR.

** we have also changed the path for saving local config to `$HOME/.config/microcks/config` from `$HOME/.microcks-cli/cofnig/`.

### Related issue(s)
Fixes #136

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,6 @@ build/dist
 
 # .DS_Store file of MacOS
 .DS_Store
+
+# test data files
+**/testdata/**
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -23,10 +23,9 @@ import (
 )
 
 var rootCmd = &cobra.Command{
-	Use:   "microcks",
-	Short: "A CLI tool for Microcks",
-	Long: `microcks-cli is a CLI for interacting with Microcks server APIs.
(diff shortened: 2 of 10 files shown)


lbroudoux (approved):
LGTM. Thanks very much!

PR #159, opened 2025-05-14
Add support for `login` command

### LINKED ISSUE #154: Add support for `login` command
### Reason/Context

- Add support for `login` command, which will help to perform authentication with microcks instance,

This is a part of #97 

### Description

- We can add `login` command to our CLI which will perform a user login and retrieve the credentials.
- It will help to perform upcoming commands seamlessly without getting credentials on every execution.
- It will also help to remove extra flags like `microcksURL`, `keycloakClietnId`, `keycloakClientSecret` from the commands.

### Implementation ideas

_No response_

Harsh4902:
### Description

- This PR adds support `login` command in our CLI tool.
- It allows users to perform `password` and `sso` based logins.
- For `password` based login, user have to set `MICROCKS_CLIENT_ID` and `MICROCKS_CLIENT_SECRET` environment variables. Then it allows users to insert username and password in interactive way as well as by `--username` and `--password` flag.
- `sso` based login will redirect the users to their default web browser, and then they can perform login operation over there. They can use `--sso-lauch-browser=false` to get authentication uri instead of getting redirected to browser.
- each execution of `login` command will insert a context for a given server and makes it a current context.

### Related issue(s)
Resolves #154


lbroudoux (approved):
Looks All Good To Me!! Excellent! 🎉

Issue #97, opened 2025-01-29
Improving Microcks CLI or creating some microcks utility like mksctl


yada:
> [!NOTE]
> This issue reflects our community's decision to submit projects to the LFX mentorship program.
> See: https://github.com/orgs/microcks/discussions/1463
> [Project Idea 1](https://github.com/orgs/microcks/discussions/1463#discussioncomment-11850083) : Improving Microcks CLI or creating some microcks utility likemksctl

We aim to improve the Microcks CLI by integrating a robust CLI framework like Cobra CLI or by developing a dedicated utility tool. This enhancement will streamline interactions with Microcks, making it easier to install, manage, and extend.

 @JulienBreux has started some work on creating a utility tool called `mksctl` [here](https://github.com/JulienBreux/mksctl), so we can take it further.

Key Goals:

- Maintain compatibility with existing test and import commands.
- Enable easy installation via package managers (brew, apt, etc.).
- Provide simple commands (mksctl start, mksctl stop) to manage Microcks instances.
- Expand functionality with new commands (e.g., import from URL, import all files in a directory, create job, list jobs).
- Ensure scalability and community contributions through standard tooling.
- Work has already begun on mksctl, and the initial implementation has been merged into the Microcks CLI 1.x branch.

Siddhi-agg:
Hi @yada !
I am Siddhi Agrawal, a fellow open-source enthusiast and I am super interested in contributing to this project for this LFX term. I have some experience in CLI tool development by contributing to my ocaml project : [ocaml-api-watch](https://github.com/ocaml-semver/ocaml-api-watch) through Outreachy. I would love to chat more about the requirements for this project. 

Do you have a pretest or any issues for this project that I can solve?
Would you want a proposal for selection of LFX mentorship?

yada:
Hi @Siddhi-agg ,

Thank you for your interest in this project; your help and contribution are greatly appreciated.

Please join as a mentee on the LFX mentorship program; this project is now online. For further details, see my latest comment in this thread: https://github.com/orgs/microcks/discussions/1463#discussioncomment-12067922

The direct link to this project is: https://mentorship.lfx.linuxfoundation.org/project/7ceac2ef-6290-4e2a-87aa-db93d909b27b
and click on "[Become a Mentee](https://mentorship.lfx.linuxfoundation.org/participate/mentee)" in the top right corner.

Then, at the beginning of March, we will select mentee(s) and move forward on the project itself with mentors.
Thank you
Regards,
[...]

PR #172, opened 2025-07-13
feat: Add `import-dir` command for bulk importing API specifications

### LINKED ISSUE #171: Add import-dir command
### Reason/Context

support for directory import from #97 

### Description

Currently, users must import API specification files individually using microcks import <file>. This is inefficient when working with directories containing multiple API files.

### Implementation ideas

_No response_

Caesarsage:
- Introduced a new command `import-dir` to allow users to import multiple API specification files from a specified directory.
- Implemented functionality to scan for supported file types (.yaml, .yml, .json, .xml) and handle recursive imports.
- Added unit tests to ensure the reliability of the import process.
- Updated documentation to include usage examples and details about the new command.

This feature enhances the CLI's capabilities for managing API specifications efficiently.

### Related issue(s)

Fixes #171 

Addresses: #97 (Support for directory import)

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -39,6 +39,7 @@ func NewCommad() *cobra.Command {
 	}
 
 	command.AddCommand(NewImportCommand(&clientOpts))
+	command.AddCommand(NewImportDirCommand(&clientOpts))
 	command.AddCommand(NewVersionCommand())
 	command.AddCommand(NewTestCommand())
 	command.AddCommand(NewImportURLCommand())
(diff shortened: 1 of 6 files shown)


Harsh4902:
@Caesarsage Thank you so much for this PR. It's actually a very good implementation for `import-dir` command. I have some concerns:
- Maybe we can use existing API sspecifications which are already present in `/sample/` directory.
- We are thinking to move all `import` related commands as sub-command like `import url`, `import dir`, so maybe we can integrate this after the migration. It will basically reduce the overhead after the new implementation.

PR #259, opened 2026-05-01
fix: read response body in DownloadArtifact

mugiwaraluffy56:
DownloadArtifact was reading from req.Body after the request completed instead of reading resp.Body. That means the CLI returned the multipart request payload, or an empty body, instead of the Microcks API response/error message from /artifact/download.

This PR changes DownloadArtifact to read resp.Body and adds an httptest coverage case that verifies the returned message comes from the server response.

Verification:
- go test ./pkg/connectors

Note: go test ./... still fails on the existing cmd/context_test.go setup because ./testdata/local.config cannot be created when cmd/testdata is absent. That failure is unrelated to this change.

### DIFF
--- a/pkg/connectors/microcks_client.go
+++ b/pkg/connectors/microcks_client.go
@@ -520,7 +520,7 @@ func (c *microcksClient) DownloadArtifact(artifactURL string, mainArtifact bool,
 	// Dump response if verbose required.
 	config.DumpResponseIfRequired("Microcks for uploading artifact", resp, true)
 
-	respBody, err := io.ReadAll(req.Body)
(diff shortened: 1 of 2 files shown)


Vaishnav88sk:
If this fixes an existing issue, please add Fixes #xxx so it auto-closes. If not, it’s good practice to create an issue before opening a PR.

mugiwaraluffy56:
> If this fixes an existing issue, please add Fixes #xxx so it auto-closes. If not, it’s good practice to create an issue before opening a PR.

sorry for the mistake, will definitely follow that from now on : )

lbroudoux (approved):
This is a very nice catch! Thanks again!

PR #271, opened 2026-05-03
fix: exit with clear error when --waitFor value is not a valid number

mugiwaraluffy56:
passing an invalid number to `--waitFor` (e.g. `--waitFor abcsec`) would silently set the timeout to 0ms. the suffix check passed because the unit was correct, but `strconv.ParseInt` returned 0 on the unparseable prefix and the error was discarded with `_`. the test then fired with a zero timeout and always failed, with no indication of what went wrong.

this also affects the invalid-suffix path: the old code printed a warning but did not exit, so the command continued with whatever value `waitForMilliseconds` happened to have.

both cases now exit with a clear message pointing at the bad value.

tested manually:
```
$ microcks test ... --waitFor abcsec
--waitFor value "abcsec" is not a valid number

$ microcks test ... --waitFor 30s
--waitFor format is wrong. Accepted units are: milli, sec, min (e.g. 500milli, 30sec, 5min)
```

### DIFF
--- a/cmd/test.go
+++ b/cmd/test.go
@@ -76,7 +76,8 @@ func NewTestCommand(globalClientOpts *connectors.ClientOptions) *cobra.Command {
 
 			// Validate presence and values of flags.
 			if !strings.HasSuffix(waitFor, "milli") && !strings.HasSuffix(waitFor, "sec") && !strings.HasSuffix(waitFor, "min") {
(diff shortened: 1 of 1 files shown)


lbroudoux (approved):
Looks legit. Thank you very much.

PR #274, opened 2026-05-04
fix: use json.Marshal in CreateResult instead of string concatenation

### LINKED ISSUE #266: fix: use json.Marshal in CreateTestResult instead of string concatenation
### Describe the bug

CreateTestResult is building JSON by hand… which is kinda fragile, so if serviceID / testEndpoint / secretName have quotes or \ → JSON can break or even allow extra fields to be injected… like http://helo.com", "injected":"val where payload gets messed up.
so its better to switch to a struct + json.Marshal (handles escaping for you)

happy to open a PR if that works!!!

rubrx:
CreateTestResult was building JSON with string concatenation, which could break or allow injection if inputs had quotes/backslashes.

Switched to a typed struct + json.Marshal for safe escaping, and properly unmarshaled filteredOperations, operationsHeaders, and oAuth2Context before combining...

Tests:
1. Malicious input test to ensure no field injection
2. Verify secretName is omitted when empty (omitempty)

fixes #266

### DIFF
--- a/pkg/connectors/microcks_client.go
+++ b/pkg/connectors/microcks_client.go
@@ -67,6 +67,18 @@ type TestResultSummary struct {
 	InProgress     bool   `json:"inProgress"`
 }
 
+// createTestResultRequest is the JSON payload sent to the Microcks tests endpoint.
+type createTestResultRequest struct {
+	ServiceID          string                 `json:"serviceId"`
+	TestEndpoint       string                 `json:"testEndpoint"`
+	RunnerType         string                 `json:"runnerType"`
+	Timeout            int64                  `json:"timeout"`
+	SecretName         string                 `json:"secretName,omitempty"`
+	FilteredOperations []string               `json:"filteredOperations,omitempty"`
+	OperationsHeaders  map[string][]HeaderDTO `json:"operationsHeaders,omitempty"`
+	OAuth2Context      *OAuth2ClientContext   `json:"oAuth2Context,omitempty"`
+}
+
 // HeaderDTO represents an operation header passed for Test
 type HeaderDTO struct {
 	Name   string `json:"name"`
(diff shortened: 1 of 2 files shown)


pkg/connectors/microcks_client.go
```diff
@@ -323,28 +335,36 @@ func (c *microcksClient) CreateTestResult(serviceID string, testEndpoint string,
...
+		ServiceID:    serviceID,
+		TestEndpoint: testEndpoint,
+		RunnerType:   runnerType,
+		Timeout:      timeout,
+		SecretName:   secretName,
 	}
 	if len(filteredOperations) > 0 && ensureValidOperationsList(filteredOperations) {
-		input += (", \"filteredOperations\": " + filteredOperations)
+		var ops []string
+		_ = json.Unmarshal([]byte(filteredOperations), &ops)
```
Vaishnav88sk, line 348: The unmarshal errors are silently discarded. If `ensureValidOperationsList` says "valid" but the value isn't a `[]string` (e.g. array of objects), the field is dropped instead of sent — a silent behavior regression vs the old code that forwarded the raw string. At minimum fall back to embedding the raw string, or return an error. I don't think we should swallow the unmarshal error.

Same concern applies to R353 and R358.

pkg/connectors/microcks_client.go
```diff
@@ -323,28 +335,36 @@ func (c *microcksClient) CreateTestResult(serviceID string, testEndpoint string,
...
+		payload.FilteredOperations = ops
 	}
 	if len(operationsHeaders) > 0 && ensureValidOperationsHeaders(operationsHeaders) {
-		input += (", \"operationsHeaders\": " + operationsHeaders)
+		var headers map[string][]HeaderDTO
+		_ = json.Unmarshal([]byte(operationsHeaders), &headers)
+		payload.OperationsHeaders = headers
 	}
 	if len(oAuth2Context) > 0 && ensureValieOAuth2Context(oAuth2Context) {
```
Vaishnav88sk, line 356: minor nit: `ensureValieOAuth2Context` has a typo ("Valie"); consider renaming while you're here. Also update wherever applicabel.

Vaishnav88sk (changes requested):
Rebase and resolve the conflicts

Caesarsage:
As per my comment here https://github.com/microcks/microcks-cli/issues/266#issuecomment-5309200229

I believe this PR can be close

PR #277, opened 2026-05-04
fix: validate start command port flag

mugiwaraluffy56:
microcks start now validates --port before it reads config or calls the container driver. Invalid values such as abc, 0, negative numbers, and values above 65535 return a clear command error instead of being passed into Docker or Podman.

Added unit coverage for valid boundary ports and invalid inputs. Tested with go test ./... and make build-binaries.

example output with an invalid port: 
image

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"fmt"
 	"log"
+	"strconv"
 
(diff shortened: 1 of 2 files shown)


Harsh4902:
@mugiwaraluffy56 This PR is very helpful. Can you please have a look at existing merge conflicts?

PR #284, opened 2026-05-05
fix: code quality improvements  vet target, bug fixes, and typo corrections

patilpratik1905:
### 1. Add `vet` target to Makefile
Added `make vet` target that runs `go vet ./...` across all packages.
Enables standardized static analysis locally and in CI pipelines.

### 2. Fix `fmt.Errorf` result not used (`pkg/watcher/executor.go`)
`fmt.Errorf(...)` was called but its return value was discarded, silently
swallowing the error. Replaced with `fmt.Printf(...)` so the error is
actually printed to output.

### 3. Fix `DownloadArtifact` reading wrong body (`pkg/connectors/microcks_client.go`)
`io.ReadAll(req.Body)` was reading the request body instead of the response
body after the HTTP call completed. This caused the function to always return
empty data. Fixed to `io.ReadAll(resp.Body)`.

### 4. Fix function name typo (`pkg/connectors/microcks_client.go`)
Renamed `ensureValieOAuth2Context` → `ensureValidOAuth2Context`.
Updated both the function definition and its call site.

## Testing
- `make vet` now passes with no errors

### Related issue(s)

NA

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,3 +28,7 @@ build-binaries:
 .PHONY: build-watcher
 build-watcher:
 	go build -o ${DIST_DIR}/${BIN_NAME}-${WATCHER_NAME} ${PACKAGE}/watcher
+
+.PHONY: vet
+vet:
+	go vet ./...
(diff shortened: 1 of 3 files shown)


Caesarsage:
The typos are covered by #300 and CI now runs go test — but the make vet idea is still good and aligns with #355. Suggest re-scoping this PR to just the vet target (or folding it into a CI-gates PR alongside staticcheck).

Harsh4902:
Thanks @Caesarsage for your review. @patilpratik1905 I'm keeping this PR open and waiting for your update.

PR #288, opened 2026-05-05
fix: return image pull stream errors

mugiwaraluffy56:
Checks the error returned while copying Docker image pull output during container creation. Before this, a read error from the pull stream was ignored and the command continued toward container creation, which could hide the real failure.

The copy logic is kept small and covered with connector tests for both success and read failure paths.

Tested with go test ./... and make build-binaries.


Caesarsage:
Thanks! The pull stream is handled on master now — DisplayJSONMessagesStream errors are returned and the stream close error is captured — so this is superseded.

I suggest closing this PR

Harsh4902:
@mugiwaraluffy56 As it's already addressed, we've to close this PR.

PR #290, opened 2026-05-05
fix: remove dead duplicate Running status check in start command

mynk8:
The `if instance.Status == "Running"` check at line 50 already returns early, making the identical `case "Running":` inside the switch statement unreachable dead code.

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -47,11 +47,6 @@ microcks start --name [name of you container/instance]`,
 				instance = &config.Instance{}
 			}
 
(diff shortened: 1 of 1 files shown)


DivyanshuVortex (approved):
LGTM, thanks for removing the duplicated `Running` check and cleaning up the unreachable logic.

Caesarsage:
Thanks — start.go was rewritten with container-state reconciliation, and the duplicated check no longer exists, so this is superseded.

This can be closed

PR #291, opened 2026-05-05
fix: parse import-url suffixes from right

### LINKED ISSUE #260: Bug: import-url loses URL path/port when parsing :primary and secret suffixes
### Describe the bug

The `import-url` command parses artifact URLs with `strings.Split(f, ":")`.

This breaks normal URLs that contain additional colons, for example URLs with ports:

```bash
http://localhost:8080/spec.yaml:true
```
The current parsing logic reconstructs the URL from only the first two split parts:
```bash
f = urlAndMainAtrifactAndSecretName[0] + ":" + urlAndMainAtrifactAndSecretName[1]
```
So http://localhost:8080/spec.yaml:true becomes http://localhost, losing the port, path, and filename.

This affects import-url when using the optional :primary and secret suffixes.

### Expected behavior

`import-url` should preserve the full URL, including scheme, host, port, path, query string, and fragment.

For example:

```bash
microcks import-url http://localhost:8080/spec.yaml:true
```
should send this URL to microcks:
```bash
http://localhost:8080/spec.yaml
```
with mainArtifact=true.

URLs without suffixes should also continue to work unchanged.

### Actual behavior

The URL is split on every colon. For a URL like:

```text
http://localhost:8080/spec.yaml:true
```
the parser splits it into parts like:
```text
["http", "//localhost", "8080/spec.yaml", "true"]
```
Then it reconstructs only:
```text
http://localhost
```
As a result, Microcks receives an incorrect artifact URL.

### How to Reproduce?

Reproducer:

1. Run `import-url` with a URL that contains a port and the `:primary` suffix:
```bash
microcks import-url http://localhost:8080/spec.yaml:true \
  --microcksURL=http://localhost:8080/api \
  --keycloakClientId=<client-id> \
  --keycloakClientSecret=<client-secret>
```

2. Observe that the CLI parses the artifact URL incorrectly before calling DownloadArtifact.

Image

Microcks CLI 1.0.2

built from source

### Additional information

A possible fix is to avoid splitting the whole URL on every colon.

Options:
- parse suffixes from the right side only, for example with `strings.LastIndex`;
- use URL-aware parsing and only treat trailing `:true`, `:false`, and optional secret values as CLI metadata;
- add focused unit tests for URLs with ports, paths, query strings, and suffixes.

Related code:

```go
urlAndMainAtrifactAndSecretName := strings.Split(f, ":")
n := len(urlAndMainAtrifactAndSecretName)
f = urlAndMainAtrifactAndSecretName[0] + ":" + urlAndMainAtrifactAndSecretName[1]
```

cotishq:
### Description

- Fix import-url argument parsing so URLs containing : (scheme/ports/etc.) are preserved when using the optional :true|false[:secret] suffixes.
- Add unit tests covering ports, paths, query strings, fragments, and secret suffix parsing.
- Update import-url docs with the new parsing behavior and examples.
- Make context_test create its testdata/ directory before writing the test config file (stabilizes test runs).

Before: 

image

After the fix:

image

### Related issue(s)

Fixes #260

### DIFF
--- a/pkg/util/artifact_specifier.go
+++ b/pkg/util/artifact_specifier.go
@@ -14,20 +14,20 @@
  * limitations under the License.
  */
 
-package cmd
+package util
 
 import (
 	"strconv"
 	"strings"
 )
 
-// parseImportFileSpecifier parses an import argument of the form:
+// ParseImportFileSpecifier parses an import argument of the form:
 //
 //	<path>[:<mainArtifactBool>]
 //
 // It parses from the right to avoid breaking paths that may contain ':'
 // characters (e.g. Windows absolute paths like C:\...).
-func parseImportFileSpecifier(spec string) (path string, mainArtifact bool) {
+func ParseImportFileSpecifier(spec string) (path string, mainArtifact bool) {
 	mainArtifact = true
 
 	lastColon := strings.LastIndex(spec, ":")
--- a/pkg/util/artifact_specifier_test.go
+++ b/pkg/util/artifact_specifier_test.go
@@ -14,13 +14,13 @@
  * limitations under the License.
  */
 
-package cmd
+package util
 
 import "testing"
 
 func TestParseImportFileSpecifier_SuffixBool(t *testing.T) {
 	in := "./specs/openapi.yaml:false"
-	path, main := parseImportFileSpecifier(in)
+	path, main := ParseImportFileSpecifier(in)
 	if path != "./specs/openapi.yaml" {
 		t.Fatalf("path mismatch: got %q", path)
 	}
@@ -31,7 +31,7 @@ func TestParseImportFileSpecifier_SuffixBool(t *testing.T) {
 
 func TestParseImportFileSpecifier_NoSuffix_Unchanged(t *testing.T) {
 	in := "./specs/openapi.yaml"
-	path, main := parseImportFileSpecifier(in)
+	path, main := ParseImportFileSpecifier(in)
 	if path != in {
 		t.Fatalf("path mismatch: got %q", path)
 	}
--- a/cmd/artifact_specifier.go
+++ b/cmd/artifact_specifier.go
@@ -1,0 +1,79 @@
+/*
+ * Copyright The Microcks Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package cmd
+
+import (
+	"strconv"
+	"strings"
+)
+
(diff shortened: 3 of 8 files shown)


cotishq:
@Harsh4902 would love your opinion on this and what changes it still requires

cotishq:
@Harsh4902 , hey can you please look into it, like what changes it still requires,

Vaishnav88sk:
@cotishq Please resolve conflicts and rebase it as overlapping with #465

Vaishnav88sk:
@Harsh4902 Can we add it to `Next` milestone?

Caesarsage:
Thanks — you had the right idea! The parse-from-the-right fix landed via #465 (parseImportURLArg), so this is superseded. Note the same class of bug still exists for file paths in import (#352) if you want a follow-up target.

Harsh4902:
@Vaishnav88sk @Caesarsage we're keeping this open and will address it in the next release after `1.0.3`.

cotishq:
@Vaishnav88sk rebased it and resolved the merge conflicts too

Harsh4902:
@cotishq You might have to make changes in you code as per the latest update in codebase and CI pipelines. I thinks that's why CI is failing.

Vaishnav88sk:
Thanks @cotishq — I re-checked against `master` after the rebase. The import-url parsing is already fixed by #465 (which added `parseImportURLArg` doing the same right-side parse). So this PR's `parseImportURLSpecifier` + the `importURL.go` changes duplicate merged code — please **drop that half** (revert `importURL.go`, remove `parseImportURLSpecifier` and its tests, and the `importURL.md` suffix section).

The part that's still valuable and missing: `parseImportFileSpecifier` + the `cmd/import.go` change (the file-path `:` bug, #352). **Keep that**, and split the `context_test.go` `MkdirAll` tweak into its own PR. Once scoped down and CI green, this is good to merge.

cotishq:
@Vaishnav88sk , @Harsh4902 Thanks for the review, i've scoped the PR down as requested: 
1) Dropped the parseImportURLSpecifier + importURL.go changes - reverted importURL.go to upstream (restoring parseImportURLArg), and removed parseImportURLSpecifier and its tests from artifact_specifier.go/test
2) Kept parseImportFileSpecifier + the cmd/import.go refactor - the right-side parse fix for the file-path : bug (#352)
3) Reverted the context_test.go MkdirAll tweak

One note: policy-guards is currently failing on master for all branches due to pkg/util/rand/rand_test.go (introduced in #494) missing the copyright header
I've opened an issue (#506)  for that seperately, please look into it.

pkg/util/artifact_specifier.go
Harsh4902, line 1: Hey @cotishq, as I can see, this file contains only utility function for parsing file specifier. I think we should not add this in `cmd` package.

pkg/util/artifact_specifier_test.go
Harsh4902, line 1: Same thing here as well. We can move this test file as well.

Harsh4902 (changes requested):

pkg/util/artifact_specifier.go
cotishq: @Harsh4902  thanks for the suggestion, i have moved artifact_specifier.go to the pkg/util, if there are some more changes to do lmk

pkg/util/artifact_specifier_test.go
cotishq: @Harsh4902 thanks for the suggestion, i have moved artifact_specifier_test.go to the pkg/util, if there are some more changes to do lmk

Issue #255, opened 2026-04-28
Feat: Microcks-CLI v2 - IDE (Vs code) Integration and Local Dev Loop Enhancement


Harsh4902:
### Reason/Context

- LFX Proposal

### Description

The Microcks CLI (microcks-cli, written in Go) is a key part of Microcks' CI/CD story - it lets developers trigger contract tests and import artifacts from the command line and in GitHub Actions. This project goes into the developer experience layer. Today, developers must navigate to the Microcks web UI in a browser to see test results, check mock status, and diagnose import errors. There is no IDE integration, no offline validation mode, and no way to get mock results inline with the code being tested. This project builds three focused improvements that bring Microcks into the editor and the local dev loop: a VS Code extension, a local dry-run mode powered by Testcontainers, and an updated GitHub Actions output format that annotates PRs with per-operation pass/fail results.

mugiwaraluffy56:
Hi @lbroudoux, @yada, @Harsh4902 - i've been working through the repository in detail to ground my proposal in the actual code. Here's my analysis and implementation plan.


### Codebase audit

After a thorough read of the repo, here are the exact integration points relevant to this project:

**Test command** (`cmd/test.go`):
The core flow (lines 48-197) POSTs to `/api/tests`, then polls `GetTestResult()` every 2 seconds with a fixed deadline. Output is plain `fmt.Printf` with no structured format or annotations. There's no `--dry-run` flag and no `--format` flag. Exit code 1 is set on test failure, which is already CI-friendly, but no per-operation annotation data is surfaced.

**Microcks client** (`pkg/connectors/microcks_client.go`):
`GetTestResult()` (line 383) returns a `*TestResultSummary` - i'll need to extend this struct (currently lines 58-69) to expose per-operation `TestCaseResult` data so github actions annotations can reference specific operation failures rather than just the top-level pass/fail.

**Container management** (`cmd/start.go`, `pkg/connectors/container_client.go`):
The CLI already wraps the docker SDK directly to `CreateContainer` / `StartContainer`. There is no testcontainers dependency and no container health-check / readiness wait. The `--dry-run` implementation will need to add `github.com/testcontainers/testcontainers-go` and the microcks module (`github.com/microcks/microcks-testcontainers-go`), then orchestrate: spin-up -> readiness check -> import -> test -> teardown - reusing the existing `MicrocksClient` interface against the ephemeral container's URL.

**Config** (`pkg/config/localconfig.go`):
[...]

PR #302, opened 2026-05-06
feat: add services list command

### LINKED ISSUE #278: feat: add services list command
The CLI can import artifacts and run tests, but it cannot currently show which services are already loaded in a Microcks instance. That makes local debugging and CI troubleshooting harder than it needs to be, because the user has to open the web UI just to confirm what is available before running a test.

Microcks already exposes service repository APIs such as `/api/services` and `/api/services/map`, so this feels like a natural CLI capability.

A possible command could be:

```bash
microcks services
# or
microcks services list
```

The default output could be a simple table with fields such as:

- service name
- version
- type
- last update or import date, if the API response exposes it cleanly

A JSON output mode could also be useful for scripts, if that matches the direction maintainers want for CLI output.

A first PR could keep the scope focused:

- add service DTOs in `pkg/connectors`
- add `MicrocksClient.GetServices`
- add a Cobra command under `cmd/`
- reuse the same auth and context resolution pattern as existing commands
- add `httptest` coverage for the client method and command output
- update command documentation

This would also be useful groundwork for the LFX VS Code integration work in #255, since listing loaded services is one of the expected editor views.

@lbroudoux @yada @Harsh4902 would this fit the CLI roadmap? If yes, I can work on a focused PR for the first version.

mugiwaraluffy56:
Closes #278.

Right now if you want to know what services are loaded in Microcks, you have to open the browser. There is no CLI equivalent of just asking "what do I have". This PR adds that.

`microcks services list` calls `GET /api/services` and prints a name/version/type table. It works the same way as `microcks test` for auth: you can either pass `--microcksURL` with keycloak credentials directly, or just run it with a saved context from `microcks login`.

Pagination is there if you need it via `--page` and `--size` (defaults to page 0, 20 per page).

I also added `ServiceSummary` and `GetServices` to the `MicrocksClient` interface so the VS Code extension work in #255 can reuse the same client method and DTO without duplicating anything.

Tests cover the happy path, empty list, pagination query params forwarded correctly, and the error path when the API returns invalid JSON. All existing tests still pass.

demo:

https://github.com/user-attachments/assets/bb619423-1cad-42a2-a75c-d2ad241f2b53

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -48,6 +48,7 @@ func NewCommad() *cobra.Command {
 	command.AddCommand(NewContextCommand(&clientOpts))
 	command.AddCommand(NewLoginCommand(&clientOpts))
 	command.AddCommand(NewLogoutCommand(&clientOpts))
+	command.AddCommand(NewServicesCommand(&clientOpts))
 
 	defaultLocalConfigPath, err := config.DefaultLocalConfigPath()
 	errors.CheckError(err)
--- a/cmd/services.go
+++ b/cmd/services.go
@@ -1,0 +1,135 @@
+/*
+ * Copyright The Microcks Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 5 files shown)


Caesarsage:
Shipped in #499 — microcks service list and microcks service get are on master (with name:version resolution and --output json|yaml). 

I believe this can be closed.

Harsh4902:
Thansk @Caesarsage for reviewing it. I am closing this.

PR #305, opened 2026-05-06
fix(watcher)!: debounce file events, bound concurrency, handle atomic saves. Closes #297

### LINKED ISSUE #297: Watcher File Change Events Trigger Unbounded Concurrent Reimports, without any Debounce or Limit
### Describe the bug

Every Write event spawns a new goroutine with no debounce, no concurrency limit, no deduplication. When a file is saved in an IDE, it typically generates 2-5 write events in rapid succession 
(write temp file, rename, metadata update). A cp or sed -i generates similar bursts. Each event spawns a separate HTTP upload to Microcks.
### Why Critical: 
1. Thundering herd: 10 files changed simultaneously leads to 10+ goroutines, 10+ concurrent uploads, potentially overwhelming the Microcks server. 
2. Race condition on the server: Two concurrent uploads of the same artifact can create conflicting service definitions, especially for large specs that take seconds to process.
3. No error aggregation: Each goroutine logs errors independently to stdout with no structured tracking. 
4. No backpressure: If the server is slow, goroutines accumulate indefinitely, this is a goroutine leak. 
5. fsnotify on directories: Some editors replace files rather than modifying in-place, which can cause the watcher to lose track of the file entirely (the inode changes).

### Expected behavior

When watching files for changes, the CLI should debounce file system events (e.g., waiting ~200-500ms after the last edit before triggering an import) to ensure only a single logical save triggers one import. It should also limit concurrency or sequence the `TriggerImport` calls to avoid overwhelming the Microcks server. Furthermore, the watcher should correctly handle atomic file saves (which often trigger `fsnotify.Rename` or `fsnotify.Create` events) so it doesn't lose track of files when they are overwritten.

### Actual behavior

Currently in [pkg/watcher/watchManager.go](pkg/watcher/watchManager.go), the `Run()` loop listens specifically for `fsnotify.Write` and immediately spawns a new, unbounded goroutine (`go TriggerImport(entry)`) for every single event.
Modern IDEs and tools (like `sed -i` or VS Code) trigger multiple rapid file system events when a file is saved (temp file write, rename, metadata update). This results in a "thundering herd" where multiple concurrent HTTP requests to upload the exact same artifact are fired at the Microcks server within milliseconds. This causes:
- Race conditions on the Microcks server resulting in potential database issues or duplicated logic.
- Goroutine leaks / absence of backpressure if the server is slow to respond.
- Loss of file tracking if the IDE performs an atomic save (replacing the file and changing its inode), because the CLI ignores non-Write operations.

### How to Reproduce?

Steps to reproduce the issue:
1. Start the microcks-cli watcher on a configuration tracking an API specification (e.g., using `microcks-cli test ...` or your dedicated watch command).
2. Open the tracked API specification file in an editor like VS Code and hit Save, or simulate rapid disk writes via terminal: `for i in {1..5}; do touch my-api-spec.yaml; done`
3. Inspect the `microcks-cli` console output. You will see multiple `[INFO] Re-importing changed file:` logs simultaneously, meaning multiple concurrent HTTP uploads for the same file are executed.

```
2026/05/05 23:29:07 [INFO] Watcher added on /root/cncf/microcks-cli/test-spec.yaml
[INFO] microcks-watcher started...
[INFO] Re-importing changed file: /root/cncf/microcks-cli/test-spec.yaml
[WARN] Error re-importing /root/cncf/microcks-cli/test-spec.yaml: Post "/test-context/api/artifact/upload": unsupported protocol scheme ""                      
[INFO] Re-importing changed file: /root/cncf/microcks-cli/test-spec.yaml
[WARN] Error re-importing /root/cncf/microcks-cli/test-spec.yaml: Post "/test-context/api/artifact/upload": unsupported protocol scheme ""                      
[INFO] Re-importing changed file: /root/cncf/microcks-cli/test-spec.yaml
[WARN] Error re-importing /root/cncf/microcks-cli/test-spec.yaml: Post "/test-context/api/artifact/upload": unsupported protocol scheme ""                      
...
```

- Microcks CLI version or git rev: `master` branch - Install method: Local build from source

_No response_

### Additional information

The root cause is in [pkg/watcher/watchManager.go](pkg/watcher/watchManager.go#L81-L91):
```go
case event := <-wm.fileWatcher.Events:
    if event.Op&fsnotify.Write == fsnotify.Write {
        // ...
        } else {
            wm.lock.Lock()
            entry, exists := wm.watchEntries[event.Name]
            wm.lock.Unlock()
            if exists {
                go TriggerImport(entry) // Spawned instantly per event
            }
        }
```

rootp1:
### Description

- Added 300ms per-file debounce to collapse rapid fsnotify event bursts (IDE saves, `sed -i`, `cp`) into a single `TriggerImport` call
- Replaced unbounded `go TriggerImport()` with a serial worker reading from a bounded `importQueue` channel (capacity 1), providing natural backpressure and preventing goroutine leaks
- Handled `fsnotify.Create`, `fsnotify.Rename`, and `fsnotify.Remove` events in addition to `fsnotify.Write`, so atomic saves (write temp + rename over target) are no longer silently missed
- Re-added the file watcher on `Create`/`Rename`/`Remove` events to track new inodes after atomic file replacements
- AddED `context.Context` propagation through `NewWatchManger`, `Run()`, `TriggerImport`, and `UploadArtifact` for clean shutdown on SIGINT/SIGTERM
- Fixed discarded `fmt.Errorf` return value bug in `executor.go` (line 16 of original)
- Added `Stop()` method to `WatchManager` for programmatic cancellation
- Added `triggerFunc` field to `WatchManager` for testability (mockable in tests)
- Added comprehensive test suite: debounce collapse, multi-file queueing, context cancellation, event filtering, atomic save handling, timer drain, `Stop()` semantics

### Reproduction results

**Before fix - Thundering herd:**
5 rapid file saves now 8-10 concurrent goroutines spawned, each firing an unbounded HTTP upload to the Microcks server.

**After fix - Debounced:**
5 rapid file saves now 1 import triggered (300ms debounce collapses all events into one).

**Before fix - Atomic save missed:**
`write temp + rename over target` emits `CHMOD` + `REMOVE` events (zero `Write` events). Old code only handled `fsnotify.Write`, so atomic saves were completely invisible.

**After fix - Atomic save handled:**
`Create`/`Rename`/`Remove` events are processed and the file watcher is re-registered on the new inode.

### Related issue(s)

Fixes thundering herd, race conditions on concurrent uploads, goroutine leaks, and lost file tracking on atomic saves as described in the watcher bug report.
Fixes #297 

### BREAKING CHANGE

`MicrocksClient.UploadArtifact` and `cmd.MicrocksClient` interface signatures now require a `context.Context` as the first parameter. Any external implementations of these interfaces must be updated accordingly.

### DIFF
--- a/cmd/import.go
+++ b/cmd/import.go
@@ -16,10 +16,13 @@
 package cmd
 
 import (
+	"context"
 	"fmt"
 	"os"
+	"os/signal"
 	"strconv"
 	"strings"
+	"syscall"
 
 	"github.com/microcks/microcks-cli/pkg/config"
 	"github.com/microcks/microcks-cli/pkg/connectors"
@@ -132,7 +135,7 @@ func NewImportCommand(globalClientOpts *connectors.ClientOptions) *cobra.Command
 				}
 
 				// Try uploading this artifact.
-				msg, err := mc.UploadArtifact(f, mainArtifact)
+				msg, err := mc.UploadArtifact(context.Background(), f, mainArtifact)
 				if err != nil {
 					fmt.Printf("Got error when invoking Microcks client importing Artifact: %s", err)
 					os.Exit(1)
--- a/cmd/import.go
+++ b/cmd/import.go
@@ -176,16 +176,19 @@ func NewImportCommand(globalClientOpts *connectors.ClientOptions) *cobra.Command
 			}
 
 			// Start watcher if --watch flag is provided.
-			if watch {
-				watchFile, err := config.DefaultLocalWatchPath()
-				errors.CheckError(err)
+	if watch {
+		watchFile, err := config.DefaultLocalWatchPath()
+		errors.CheckError(err)
 
-				wm, err := watcher.NewWatchManger(watchFile)
-				errors.CheckError(err)
+		ctx, stop := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
+		defer stop()
 
-				fmt.Println("Watch mode enabled - microcks-watcher started...")
-				wm.Run()
-			}
+		wm, err := watcher.NewWatchManger(ctx, watchFile)
+		errors.CheckError(err)
+
+		fmt.Println("Watch mode enabled - microcks-watcher started...")
+		wm.Run()
+	}
 		},
 	}
 
--- a/cmd/importDir.go
+++ b/cmd/importDir.go
@@ -16,6 +16,7 @@
 package cmd
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -28,7 +29,7 @@ import (
 
 // MicrocksClient interface for dependency injection
 type MicrocksClient interface {
-	UploadArtifact(file string, main bool) (string, error)
+	UploadArtifact(ctx context.Context, file string, main bool) (string, error)
 }
 
 type FileType struct {
@@ -231,7 +232,7 @@ func ImportDirectory(client MicrocksClient, fs FileSystem, dirPath string, confi
 	for _, file := range files {
 		fileType := detectFileType(file)
 
-		msg, err := client.UploadArtifact(file, fileType.IsPrimary)
+		msg, err := client.UploadArtifact(context.Background(), file, fileType.IsPrimary)
 		if err != nil {
 			result.FailedCount++
 			result.FailedFiles = append(result.FailedFiles, file)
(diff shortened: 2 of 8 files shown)


rootp1:
@Harsh4902, @yada, @lbroudoux 
Here is my work for the issue #297
I have tried to align with most of the contribution guidelines, Please review this PR whenever you get a chance to. Thankyou

PR #307, opened 2026-05-06
fix : `start` and `stop` commands crash when container state diverges from local config

### LINKED ISSUE #306: fix : `start` and `stop` commands crash when container state diverges from local config
### Reason/Context

The CLI currently crashes if you manually delete a container (using `docker rm` or `prune`) because it only looks at its own config file, not the actual Docker/Podman state. This fix makes the CLI smarter by checking the real state of your containers before acting. 

This prevents users from getting "stuck" in a state where the CLI crashes and won't let you start or stop until you manually delete your config file.

**Reproduction:**
```bash
microcks start
docker rm -f microcks
microcks start # False Output: "Microcks instance is already running"
```
### Description

The CLI needs to ask the Docker daemon for the real container state *before* acting on the state stored in the config file. 
Changes introduced:
1. Check the Docker daemon for container existence before attempting to create or stop it. 
2. Update `UpsertInstance` in `localconfig.go` to match by `Name` instead of `ContainerID`. Currently, every time a container is recreated (getting a new ID), it appends a duplicate ghost entry to the config file instead of updating the existing one.

### Implementation ideas

Add a `GetContainer(name string)` method to the `ContainerClient` interface to inspect real container state

DivyanshuVortex:
## Description : 
------

- Fixed `microcks start` to sync with daemon state before starting containers
- Fixed `microcks stop` to handle already-removed containers gracefully
- Fixed `UpsertInstance` to match by `Name` instead of `ContainerID`
- Prevented duplicate ghostt entries in `~/.config/microcks/config`
- Added `GetContainer(name)` to inspect real container state from Docker/Podman
----
Screenshot (685)

-----

Resolves : #306

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -47,30 +47,34 @@ microcks start --name [name of you container/instance]`,
 				instance = &config.Instance{}
 			}
 
-			if instance.Status == "Running" {
-				fmt.Printf("Microcks instance with name %s is already running", name)
-				return
+			containerClient, err := connectors.NewContainerClient(driver)
+			errors.CheckError(err)
+			defer containerClient.CloseClient()
+
+			daemonID, daemonStatus, err := containerClient.GetContainer(name)
+			if err != nil {
+				log.Fatalf("failed to inspect container: %v", err)
 			}
 
-			switch instance.Status {
-			case "Running":
-				fmt.Printf("Microcks instance with name %s is already running", name)
-				return
-			case "Exited":
-				containerClient, err := connectors.NewContainerClient(instance.Driver)
-				errors.CheckError(err)
-				defer containerClient.CloseClient()
-
-				if err := containerClient.StartContainer(instance.ContainerID); err != nil {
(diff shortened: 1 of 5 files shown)


DivyanshuVortex:
Removed `return` statemnts following log.Fatalf(). bcoz, log.Fatalf calls os.Exit(1) and terminates the program immediately, making any subsequent return unreachable.

rootp1:
@DivyanshuVortex, how do you temporarily deploy your PR to build?

DivyanshuVortex:
> @DivyanshuVortex, how do you temporarily deploy your PR to build?

GitHub Actions is set to auto-run on PR , so the temp Build deploy just kicked off automatically when the PR got opened/updated.

DivyanshuVortex:
Hey @Harsh4902, could you please review this PR??? thnx

Syedowais312:
This is fixed in this PR https://github.com/microcks/microcks-cli/pull/456

PR #311, opened 2026-05-06
feat: add --wait flag to block until Microcks is ready after start

### LINKED ISSUE #303: feat: wait for Microcks readiness after start
`microcks start` returns as soon as the container is created, but the Microcks server inside takes a few seconds to actually come up. So if you immediately run `microcks import` or `microcks test` after starting, you get a connection refused error and have to either sleep or retry manually.

```bash
microcks start
microcks import myapi.yaml    # connection refused
```
image1

image2

The Microcks Testcontainers module already handles this by polling `GET /api/health` before handing control back to the caller. Would be good to have the same in `microcks start` via a `--wait` flag, optionally with a `--wait-timeout`.

```bash
microcks start --wait
# blocks until /api/health returns 200
```

This also ties into the `--dry-run` work in #255, where the ephemeral container needs to be ready before importing and testing.

@lbroudoux @yada @Harsh4902 would a `--wait` flag work here, or would you prefer readiness wait to be always-on?

mugiwaraluffy56:
Closes #303.

`microcks start` returns as soon as the container is created, but the server inside takes a few seconds to initialize. Running `microcks import` or `microcks test` right after start fails with a connection error because the API is not up yet.

Adding `--wait` makes `microcks start` block and poll `GET /api/health` on the mapped port until it returns 200, then exit. An optional `--wait-timeout` flag (default 30 seconds) controls how long to wait before giving up with an error.

```bash
microcks start --wait
microcks start --wait --wait-timeout 60
```

This is the same readiness strategy the Microcks Testcontainers module uses internally. It also makes the `--dry-run` path in #255 straightforward to build on top of, since it needs to know when the ephemeral container is ready before importing an artifact.

**Before (no --wait):**
bug

**After (with --wait):**
wait

**Import succeeds after --wait:**
import

Tests cover the happy path, retry-then-succeed, timeout, and URL construction. All existing tests still pass.

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -3,6 +3,9 @@ package cmd
 import (
 	"fmt"
 	"log"
+	"net/http"
+	"os"
+	"time"
 
 	"github.com/microcks/microcks-cli/pkg/config"
 	"github.com/microcks/microcks-cli/pkg/connectors"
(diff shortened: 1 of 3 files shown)


Caesarsage:
Thank you @mugiwaraluffy56 

This is built into start on master — it now polls the server until HTTP is actually answering and exposes --ready-timeout (default 60s), so chained import/test commands don't race the boot. 

I believe this can be closed.

Harsh4902:
As this feature is already shipped, we've to close this.

PR #315, opened 2026-05-07
improve TestDeleteContext using temporary directories

Manar-Elhabbal7:
### Description
This PR fixes a brittle test case in `cmd/context_test.go` where **TestDeleteContext** was failing on
  non-Unix-native filesystems (like **NTFS/FAT**). 

  The test previously relied on a static file path in ./testdata/, which caused issues when trying to set
  and verify restrictive file permissions (0600). By switching to t.TempDir(), the test is now isolated,
  environment-agnostic, and leaves no garbage files behind.

###   Changes
   - Removed hardcoded testConfigFilePath constant.
   - Added path/filepath to imports.
   - Updated TestDeleteContext to generate a unique temporary directory and path for each execution.

### DIFF
--- a/cmd/context_test.go
+++ b/cmd/context_test.go
@@ -3,7 +3,7 @@ package cmd
 import (
 	"os"
 	"testing"
-
+	"path/filepath"
 	"github.com/microcks/microcks-cli/pkg/config"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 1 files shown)


Caesarsage:
Thanks @Manar-Elhabbal7 

This has largely being addressed on master, the newer context tests use t.TempDir() and the remaining fixture creates its directory explicitly, so the clean-clone failure is gone. If a specific NTFS/FAT permission failure remains, a fresh repro against master would help; otherwise I believe this can be closed.

PR #317, opened 2026-05-07
fix(login): use isolated HTTP mux, eagerly bind port, and return errors from oauth2login. Fixes #312

### LINKED ISSUE #312: Auth2 SSO Callback Server Is a Global Singleton, Race Condition & Port Hijack
### Describe the bug

The OAuth2 SSO Callback Server in the CLI (used during `login --sso`) acts as a global singleton due to its reliance on `http.DefaultServeMux`. This leads to race conditions and potential port hijacking vulnerabilities.

In `cmd/login.go:268-270` (approximate location):
```go
srv := &http.Server{Addr: "localhost:" + strconv.Itoa(port)}
http.HandleFunc("/auth/callback", callbackHandler)
```
This logic instantiates an `http.Server` but does not supply a dedicated `Handler`, meaning it falls back to the process-global `http.DefaultServeMux`. `http.HandleFunc` modifies this shared router, and this route is never cleaned up.

The critical impacts include:
1. **Port Hijacking / Token Exposure**: If a malicious process running locally on a multi-tenant machine binds to port 58085 first, the CLI's temporary HTTP server will fail to start (`log.Fatalf`), but the callback/authorization details from the user's browser may be successfully routed to the attacker. While PKCE prevents the attacker from directly exchanging the authorization code (as the `code_verifier` is in the memory of the CLI process), the `id_token` (passed as `r.FormValue("id_token")` in the fallback implicit flow) might be extracted directly by the attacking server.
2. **Global Mux Pollution & Race Conditions**: The URL handler `/auth/callback` persists permanently inside the process's default HTTP mux. If `login` is called twice (e.g., in a CI script, or if the first login fails midway), the subsequent calls overwrite the first handler. `http.HandleFunc` is not completely goroutine-safe when dynamically modifying the mux while a server handles requests. It can incorrectly interleave or fail across consecutive executions in the same process footprint.

### Expected behavior

1. A locally isolated `http.ServeMux` should be used per server instance instead of `http.DefaultServeMux`. For example:
   ```go
   mux := http.NewServeMux()
   mux.HandleFunc("/auth/callback", callbackHandler)
   srv := &http.Server{
       Addr: "localhost:" + strconv.Itoa(port),
       Handler: mux,
   }
   ```
2. The port binding error logic should be explicit: if the expected local port is already bound by another process, `login()` should abort safely and alert the user immediately, rather than quietly failing inside a goroutine or leaving the user vulnerable to authorization route hijacking.

### Actual behavior

The `http.HandleFunc` manipulates the global `http.DefaultServeMux`. The CLI process could leak handlers or have subsequent authentication operations overwrite global states over one another.

### How to Reproduce?

1. Open a separate terminal and mock an attacker process by listening on port 58085:
   ```bash
   nc -l -p 58085
   ```
2. Run the `microcks-cli login --sso` command.
3. Observe that the OAuth authentication flow continues locally.
4. When the browser redirects to `http://localhost:58085/auth/callback...`, the attacker `nc` instance intercepts the HTTP request, collecting the query parameters (which may contain `code` or `id_token` depending on the SSO provider mapping configuration).

Alternatively, inspecting the source code validates the reliance on `http.DefaultServeMux`:
1. Find `srv := &http.Server{Addr: "localhost:" + strconv.Itoa(port)}` inside `cmd/login.go`.
2. Notice that `http.HandleFunc` is being used, making `/auth/callback` globally registered on `http.DefaultServeMux`.

microcks-cli2 (Current main/issue-branch code base as of reported date)

Compiled from source (`go build`)

### Additional information

This issue affects multi-tenant environments where CI/CD setups or shared developer workstations might have concurrent or conflicting listeners. Fixing the HTTP handler scoping provides a resilient boundary.

rootp1:
### Description

- Replaced `http.HandleFunc` (process-global `http.DefaultServeMux`) with an isolated `http.NewServeMux` per `oauth2login` call, eliminating handler pollution across repeated invocations and race conditions on the shared global mux
- Changed `oauth2login` to eagerly bind the callback port via `net.Listen` before opening the browser, so port hijack attempts are detected immediately with a clear error message instead of crashing inside a goroutine after the OAuth flow has already started
- Reordered server startup: the callback HTTP server now starts before `ssoAuthFlow` opens the browser, guaranteeing the server is ready when the browser redirects back
- Changed `oauth2login` signature from `(string, string)` to `(string, string, error)`, replacing `log.Fatal`/`log.Fatalf` calls with returned errors so the caller controls error handling and the function is testable
- Added `srv.Shutdown` on the error path before returning, ensuring the callback server port is released even when the OAuth flow fails
- Removed the `time.Sleep(1 * time.Second)` hack that previously tried to paper over the race between server startup and browser redirect

### Reproduction results

**Before fix - Port hijack:**
A rogue process listening on port 58085 before `login --sso` causes `log.Fatalf` inside a goroutine after the browser is already open. The user sees a confusing crash and the OAuth callback is silently routed to the attacker.

**After fix - Port hijack detected early:**
`oauth2login` returns `error: "failed to bind callback port localhost:58085: listen tcp 127.0.0.1:58085: bind: address already in use (is another process using it?)"` immediately - no browser opened, no OAuth flow started.

**Before fix - Global mux pollution:**
`http.HandleFunc("/auth/callback", ...)` registers the route on `http.DefaultServeMux` permanently. A second `login --sso` call overwrites the first handler. In the same binary, this route interferes with any other HTTP serving.

**After fix - Isolated mux:**
Each `oauth2login` call creates its own `http.NewServeMux`. `/auth/callback` is never registered on `http.DefaultServeMux`. Concurrent and sequential calls are fully independent.

**Before fix - Server port leaked on error:**
When the callback receives an error (e.g. `access_denied`), `log.Fatal` is called without shutting down the server, leaving the port bound until the process exits.

**After fix - Clean shutdown on error:**
`srv.Shutdown` is called before returning the error, releasing the port immediately.

**Test suite (5 tests, all passing):**

| Test | Verifies |
|------|----------|
| `TestPortHijackDetection` | Rogue listener → clear early abort with descriptive error |
| `TestGlobalMuxNotPolluted` | `/auth/callback` not in `DefaultServeMux` after call |
| `TestEagerPortBindBeforeBrowser` | Port bound before browser would open |
| `TestServerShutdownOnErrorPath` | Port released after error callback |
| `TestConcurrentIsolatedMuxCalls` | Two concurrent calls bind independent ports |

### Related issue(s)

Fixes race condition, global mux pollution, and port hijack vulnerability in OAuth2 SSO callback server as described in issue 
Fixes #312 .

### BREAKING CHANGE

`oauth2login` now returns `(string, string, error)` instead of `(string, string)`. Any callers of this function must handle the returned error.

### DIFF
--- a/cmd/login.go
+++ b/cmd/login.go
@@ -266,8 +266,12 @@ func oauth2login(
 		completionChan <- ""
 	}
 
-	srv := &http.Server{Addr: "localhost:" + strconv.Itoa(port)}
-	http.HandleFunc("/auth/callback", callbackHandler)
+	mux := http.NewServeMux()
+	mux.HandleFunc("/auth/callback", callbackHandler)
+	srv := &http.Server{
+		Addr:    "localhost:" + strconv.Itoa(port),
+		Handler: mux,
+	}
 
 	var url string
 	opts := []oauth2.AuthCodeOption{}
--- a/cmd/login.go
+++ b/cmd/login.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"html"
 	"log"
+	"net"
 	"net/http"
 	"os"
 	"strconv"
--- a/cmd/login.go
+++ b/cmd/login.go
@@ -286,15 +286,15 @@ func oauth2login(
 		return "", "", fmt.Errorf("failed to bind callback port %s: %v (is another process using it?)", srv.Addr, listenErr)
 	}
 
-	fmt.Printf("Performing %s flow login: %s\n", "authorization_code", url)
-	time.Sleep(1 * time.Second)
-	ssoAuthFlow(url, ssoLaunchBrowser)
 	go func() {
 		log.Printf("Listen: %s\n", srv.Addr)
 		if serveErr := srv.Serve(ln); serveErr != nil && serveErr != http.ErrServerClosed {
 			completionChan <- fmt.Sprintf("Temporary HTTP server failed: %s", serveErr)
 		}
 	}()
+
+	fmt.Printf("Performing %s flow login: %s\n", "authorization_code", url)
+	ssoAuthFlow(url, ssoLaunchBrowser)
 	errMsg := <-completionChan
 	if errMsg != "" {
 		log.Fatal(errMsg)
--- a/cmd/login.go
+++ b/cmd/login.go
@@ -297,7 +297,10 @@ func oauth2login(
 	ssoAuthFlow(url, ssoLaunchBrowser)
 	errMsg := <-completionChan
 	if errMsg != "" {
-		log.Fatal(errMsg)
+		shutdownCtx, shutdownCancel := context.WithTimeout(context.Background(), 1*time.Second)
+		defer shutdownCancel()
+		_ = srv.Shutdown(shutdownCtx)
+		return "", "", fmt.Errorf("%s", errMsg)
 	}
 	fmt.Printf("Authentication successful\n")
 	ctx, cancel := context.WithTimeout(ctx, 1*time.Second)
--- a/cmd/login.go
+++ b/cmd/login.go
@@ -189,7 +189,7 @@ func oauth2login(
 	port int,
 	oauth2conf *oauth2.Config,
 	ssoLaunchBrowser bool,
-) (string, string) {
+) (string, string, error) {
 	oauth2conf.ClientID = "microcks-app-js"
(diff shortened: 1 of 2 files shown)


rootp1:
> `oauth2login` now returns `(string, string, error)` instead of `(string, string)`. Any callers of this function must handle the returned error.

Hey @Harsh4902, @lbroudoux @yada,
It would be great to know your views on this change

PR #319, opened 2026-05-07
fix: return errors instead of panicking on response parsing in connectors

### LINKED ISSUE #318: Replace panic() with error returns in connectors package
### Bug description
Several methods in pkg/connectors/microcks_client.go and pkg/connectors/keycloak_client.go call panic() when io.ReadAll or json.Unmarshal fail on an HTTP response. Because every one of these methods already has error in its return signature, panicking instead of returning the error means any malformed/truncated response from the Microcks server or Keycloak crashes the entire CLI with a stack trace, instead of producing a clean user-facing error.

This is the same anti-pattern that was previously fixed in:

[#258](https://github.com/microcks/microcks-cli/pull/258) (json.Unmarshal ignored in GetTestResult)
[#259](https://github.com/microcks/microcks-cli/pull/259) (DownloadArtifact not reading body)
f9f282c (replaced one panic in GetTestResult with a wrapped error)
The recent f9f282c only fixed a single occurrence. The same panic(err) / panic(err.Error()) pattern still exists in 11 other places.

### Reproduction
Any scenario where the Microcks or Keycloak server returns a non-JSON body (HTML error page from a reverse proxy, truncated body on connection drop, gateway timeout HTML, etc.) on an otherwise-2xx-coded request will crash the CLI with a Go panic stack trace instead of a friendly error.

### Proposed fix
Replace each panic(err) / panic(err.Error()) with a wrapped fmt.Errorf("...: %w", err) return, matching the style of the recent fix at microcks_client.go:413-415.

Heyy-Himanshuu:
### Summary
Replaces 12 panic(err) / panic(err.Error()) calls in pkg/connectors/{microcks,keycloak}_client.go with wrapped fmt.Errorf("...: %w", err) returns.
Cleans up three return x, err sites where err was guaranteed nil after the change (now return x, nil).
Continues the pattern established by #258, #259 and commit f9f282c, which fixed the same anti-pattern in one location each.
fixes #318 

### Why
Every affected method already returns error. Panicking on io.ReadAll / json.Unmarshal / io.Copy failure means a malformed or truncated response from Microcks/Keycloak crashes the CLI with a stack trace instead of surfacing a clean error. This is most visible when a reverse proxy or gateway returns HTML on an otherwise-2xx request, or when the connection drops mid-body.

### Test plan
 

- [x] go build ./... passes
- [x]  go test ./pkg/connectors/... passes (existing TestDownloadArtifactReturnsResponseBody still green)
- [ ]  Manual: point CLI at a server returning HTML on /api/keycloak/config — previously panicked, now returns a wrapped error

### DIFF
--- a/pkg/connectors/keycloak_client.go
+++ b/pkg/connectors/keycloak_client.go
@@ -97,16 +97,16 @@ func (c *keycloakClient) ConnectAndGetToken() (string, error) {
 
 	body, err := io.ReadAll(resp.Body)
 	if err != nil {
-		panic(err.Error())
+		return "", fmt.Errorf("failed to read Keycloak token response: %w", err)
 	}
 
 	var openIDResp map[string]interface{}
 	if err := json.Unmarshal(body, &openIDResp); err != nil {
-		panic(err)
+		return "", fmt.Errorf("failed to parse Keycloak token response: %w", err)
 	}
 
 	accessToken := openIDResp["access_token"].(string)
-	return accessToken, err
+	return accessToken, nil
 }
 
 func (c *keycloakClient) GetOIDCConfig() (*oauth2.Config, error) {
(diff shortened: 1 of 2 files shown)


Caesarsage:
as per my comments here: https://github.com/microcks/microcks-cli/issues/318#issuecomment-5309165687

I believe this can be closed

PR #320, opened 2026-05-07
feat: add tests list command to show test history for a service

### LINKED ISSUE #310: feat: add tests list command to show test history for a service
After running `microcks test`, the CLI exits with a pass/fail result and a link to the web UI. If you want to look at the history of test runs for a service (how many ran, when, which endpoint was tested), you still have to open the browser.

Microcks already exposes this via `GET /api/tests?serviceId=<name:version>`, so the data is there. It just needs a CLI surface.

A simple command like:

```bash
microcks tests list "WeatherForecast API:1.1.0"
```

could print a table showing the test ID, date, runner type, tested endpoint, and pass/fail status for each run.

This makes it easier to debug flaky test runs, audit what was tested and when, and confirm that a recent import actually triggered a new test, all without leaving the terminal.

It also ties into the VS Code extension work in #255, where showing recent test results per service is one of the expected views.

@lbroudoux @yada @Harsh4902 does this fit the CLI roadmap? Happy to open a PR once there is a thumbs up on the direction.

mugiwaraluffy56:
Closes #310.

Adds a `microcks tests list <serviceRef>` command that prints a table of test results for a given service, including test ID, date, runner type, tested endpoint, and pass/fail status. Supports `--page` and `--size` flags for pagination.

The implementation follows the same pattern as the existing commands: `GetTestResults(serviceRef string, page, size int)` added to the `MicrocksClient` interface and implementation, calling `GET /api/tests?serviceId=<ref>&page=<n>&size=<n>`. Also adds `RunnerType` to `TestResultSummary` since the field is returned by the API but was not previously mapped.

Four `httptest`-based tests cover the normal case, empty results, pagination params, and invalid JSON response.

microcks tests list output

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -48,6 +48,7 @@ func NewCommad() *cobra.Command {
 	command.AddCommand(NewContextCommand(&clientOpts))
 	command.AddCommand(NewLoginCommand(&clientOpts))
 	command.AddCommand(NewLogoutCommand(&clientOpts))
+	command.AddCommand(NewTestsCommand(&clientOpts))
 
 	defaultLocalConfigPath, err := config.DefaultLocalConfigPath()
 	errors.CheckError(err)
--- a/cmd/tests.go
+++ b/cmd/tests.go
@@ -1,0 +1,130 @@
+/*
+ * Copyright The Microcks Authors.
+ *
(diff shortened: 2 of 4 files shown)


Caesarsage:
Shipped in #499 — microcks test list <serviceRef> and microcks test get <id> exist on master with --output json|yaml support. 

I believe this can be closed.

Issue #321, opened 2026-05-07
TestDeleteContext fails when cmd/testdata directory is missing


hazelmayank:
### Describe the bug

While running the test suite with `go test ./...`, the `TestDeleteContext` test fails because it tries to write/read a config file at `./testdata/local.config`. However, the `cmd/testdata` directory does not exist in a fresh clone, causing the test to fail with:

```
open ./testdata/local.config: no such file or directory
```

This makes the test dependent on a manually existing local directory instead of being fully self-contained.

<img width="831" height="848" alt="Image" src="https://github.com/user-attachments/assets/a54eb86e-1209-48c7-a464-612409527840" />

### Expected behavior

`go test ./...` should pass on a fresh clone without requiring the `cmd/testdata` directory to already exist. The test should create any temporary files or directories it needs during execution.

### Actual behavior

The test suite fails in `TestDeleteContext`:

```
--- FAIL: TestDeleteContext (0.00s)
    context_test.go:44:
        Error: Received unexpected error:
        open ./testdata/local.config: no such file or directory
```

### How to Reproduce?

1. Clone the repository fresh: `git clone https://github.com/microcks/microcks-cli.git`
2. Change into the directory: `cd microcks-cli`
3. Run the test suite: `go test ./...`
4. Observe the failure in `cmd/context_test.go` at `TestDeleteContext`.

### Additional information

**Root cause.** `cmd/context_test.go` declares a hardcoded path:

```go
const testConfigFilePath = "./testdata/local.config"
```

and then calls `os.WriteFile(testConfigFilePath, ...)`. `os.WriteFile` does not create parent directories, and `.gitignore` excludes `**/testdata/**`, so the `cmd/testdata/` directory does not exist on a clean clone.

**Proposed fix.** Use Go's `t.TempDir()` so the test creates an isolated, auto-cleaned temporary directory on every run:

```go
testConfigFilePath := filepath.Join(t.TempDir(), "local.config")
```

This makes the test self-contained, removes the dependency on repository-local folders, and prevents state leakage between consecutive runs.

I would like to work on this issue if the maintainers agree with the proposed approach. I can send a small PR that updates the test to use `t.TempDir()`.

hazelmayank:
I reproduced this on a clean upstream master.

Summary:
`TestDeleteContext` writes to `./testdata/local.config`, but `cmd/testdata` does not exist on a clean clone. Since `os.WriteFile` does not create parent directories, the test fails with:

[...]

PR #334, opened 2026-05-08
test: use temporary config file in context delete test

### LINKED ISSUE #313: Refactor :  improve TestDeleteContext robustness and isolation using temporary directories
### Reason/Context

  Currently, `TestDeleteContext` in `cmd/context_test.go` relies on a **hardcoded** file path
  `/testdata/local.config`

  This approach causes the test to fail on filesystems that do not fully
  support Unix-style permissions (such as NTFS or FAT32 partitions commonly used on external drives or
  Windows-mounted volumes), as the test attempts to verify restrictive file permissions (0600).

### Description

The enhancement involves refactoring the `TestDeleteContext` function. Instead of using a static file
  path, the test will now:

   1. Utilize `t.TempDir()` to create a unique, isolated directory for each test run.
   2. Construct the configuration file path dynamically using f`ilepath.Join()`
   3. Clean up automatically after the test finishes, as `t.TempDir()` is managed by the Go testing
      package.

### Implementation ideas

_No response_

hazelmayank:
## Summary

This PR fixes `TestDeleteContext` by updating it to use a temporary config file created with `t.TempDir()` instead of writing to `./testdata/local.config`.

On a clean checkout, `cmd/testdata` may not exist, which causes the test to fail with:

open ./testdata/local.config: no such file or directory

The test behavior remains unchanged, but it no longer depends on an untracked local directory.

Fixes #313
Related to #321

## Test plan
 go test ./cmd/... -run TestDeleteContext -v
 go test ./...
 go test ./cmd/... -count=3
 make build-local

### DIFF
--- a/cmd/context_test.go
+++ b/cmd/context_test.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"os"
+	"path/filepath"
 	"testing"
 
 	"github.com/microcks/microcks-cli/pkg/config"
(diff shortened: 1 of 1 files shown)


Caesarsage:
thanks @hazelmayank 

This no longer reproducible on master (same report as #417) — the fixture directory is created by the test itself now. 

I believe this can be closed.

PR #338, opened 2026-05-08
fix typo in DownloadArtifact: was reading req.Body instead of resp.Body

### LINKED ISSUE #337: Bug: DownloadArtifact reads from req.Body instead of resp.Body
## Body

In `pkg/connectors/microcks_client.go:523`, `DownloadArtifact` reads from `req.Body` after the HTTP call returns:

```go
respBody, err := io.ReadAll(req.Body)
```

`req.Body` is the outgoing request body (the multipart form). It's already consumed by `httpClient.Do(req)`, so `io.ReadAll` here returns 0 bytes. I confirmed with a small `httptest` reproduction - reading `req.Body` gave an empty slice; reading `resp.Body` gave the actual server response.

### Effect on microcks import-url

- on HTTP 201 it prints `Microcks has discovered ''` (empty quotes, no confirmation of what was ingested)
- on a non-201 response the returned error has no message - the CLI prints `Got error when invoking Microcks client importing Artifact:` with nothing after it and exits 1, so CI fails with no diagnostic

### Fix

change `req.Body` to `resp.Body` on that line.

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

_No response_

_No response_

### Additional information

_No response_

Priyanshubhartistm:
After `httpClient.Do(req)` returns, `req.Body` is already consumed - so `io.ReadAll(req.Body)` always gives back empty bytes. Swapped it to `resp.Body` so we actually get the server's response back.

Also fixed the two log labels that still said "uploading artifact" inside the download path (copy-paste leftover from `UploadArtifact`).

### Description

- `DownloadArtifact` was calling `io.ReadAll(req.Body)` instead of `io.ReadAll(resp.Body)`, meaning the server response was always silently discarded
- On a 201 the CLI printed `Microcks has discovered ''` (empty string - service name was never captured)
- On any non-201 the error message was blank, so CI would fail with `Got error when invoking Microcks client importing Artifact:` and no further detail
- Fixed the two `DumpRequestIfRequired` / `DumpResponseIfRequired` log labels in `DownloadArtifact` that still said `"uploading artifact"` (copy-pasted from `UploadArtifact` and never updated)

### Related issue(s)

Fixes #337

### DIFF
--- a/pkg/connectors/microcks_client.go
+++ b/pkg/connectors/microcks_client.go
@@ -510,7 +510,7 @@ func (c *microcksClient) DownloadArtifact(artifactURL string, mainArtifact bool,
 	req.Header.Set("Authorization", "Bearer "+c.AuthToken)
 
 	// Dump request if verbose required.
-	config.DumpRequestIfRequired("Microcks for uploading artifact", req, true)
+	config.DumpRequestIfRequired("Microcks for downloading artifact", req, true)
 
 	resp, err := c.httpClient.Do(req)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


Priyanshubhartistm:
Hi @Harsh4902, @lbroudoux

Could you please review this PR when you get a chance?

Caesarsage:
As per my comments here

https://github.com/microcks/microcks-cli/issues/337#issuecomment-5309184402

I believe this can be closed

PR #341, opened 2026-05-08
fix(connectors): use json.Marshal in CreateTestResult to prevent injection and panics

### LINKED ISSUE #340: fix(connectors): CreateTestResult builds JSON via string concat — field injection and panic risks
### Describe the bug

CreateTestResult in pkg/connectors/microcks_client.go builds the POST /tests request body by concatenating raw strings without any JSON escaping. This means:
1. An API name containing a double-quote or backslash produces malformed JSON, causing an opaque 400 from the server with no actionable error message.
2. A crafted serviceRef value containing quote characters can inject extra fields into the JSON object sent to the server — a JSON injection vulnerability relevant in CI/CD pipelines where these values come from environment variables or PR branch names.
3. The same function calls panic() on body read failure and on JSON unmarshal failure instead of returning errors, and uses an unchecked type assertion createTestResp["id"].(string) that panics if the server returns an unexpected shape.

### Expected behavior

Special characters in serviceID, testEndpoint, and secretName are safely escaped. Errors in reading or parsing the response are returned as errors, not panics.

### Actual behavior

Raw string concatenation is used:
  input += ("\"serviceId\": \"" + serviceID + "\", ")
  input += ("\"testEndpoint\": \"" + testEndpoint + "\", ")
A value like:  foo", "timeout": 999999, "x": "  silently restructures the JSON body. panic() is called on read/unmarshal errors instead of returning them.

### How to Reproduce?

1. Run: microcks test 'My "API":1.0' http://endpoint OPEN_API_SCHEMA --waitFor 5sec
2. Observe malformed JSON error or unexpected server 400 response
3. Alternatively inspect the raw POST body — the double-quote in the API name breaks JSON syntax

microcks-cli v1.0.3 — git rev b892bbb (upstream/master)

this is a CLI client-side bug in pkg/connectors/microcks_client.go lines 327–345

### Additional information

The three "complex" parameters (filteredOperations, operationsHeaders, oAuth2Context) already go through json.Unmarshal validation before being embedded, proving the correct approach is known. The fix is to use a typed struct with json.Marshal for the entire body. A PR with the fix is ready.

Harishrs2006:
Replace hand-rolled string concatenation in `CreateTestResult` with a typed struct and `json.Marshal` — special characters in `serviceID`, `testEndpoint`, and `secretName` are now safely escaped
1) Eliminates JSON field injection risk when values come from environment variables or external sources in CI/CD pipelines
2) Replace three `panic()` calls with proper `error` returns
3) Add safe checked type assertion for the response `id` field instead of bare `.(string)` that panics on unexpected server response

Related issues: Fixes #340

### DIFF
--- a/pkg/connectors/microcks_client.go
+++ b/pkg/connectors/microcks_client.go
@@ -318,33 +318,52 @@ func (c *microcksClient) SetOAuthToken(oauthToken string) {
 	c.AuthToken = oauthToken
 }
 
+// testResultRequest is the serializable body for the POST /tests endpoint.
+type testResultRequest struct {
+	ServiceID          string           `json:"serviceId"`
+	TestEndpoint       string           `json:"testEndpoint"`
+	RunnerType         string           `json:"runnerType"`
+	Timeout            int64            `json:"timeout"`
+	SecretName         string           `json:"secretName,omitempty"`
+	FilteredOperations *json.RawMessage `json:"filteredOperations,omitempty"`
+	OperationsHeaders  *json.RawMessage `json:"operationsHeaders,omitempty"`
+	OAuth2Context      *json.RawMessage `json:"oAuth2Context,omitempty"`
+}
+
 func (c *microcksClient) CreateTestResult(serviceID string, testEndpoint string, runnerType string, secretName string, timeout int64, filteredOperations string, operationsHeaders string, oAuth2Context string) (string, error) {
 	// Ensure we have a correct URL.
 	rel := &url.URL{Path: "tests"}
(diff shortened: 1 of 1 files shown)


Caesarsage:
Thanks for this! The typed-struct + json.Marshal approach (including validation of filteredOperations/headers/oauth sub-fields) landed on master, so this PR is superseded. The injection analysis in the description was on point.

I suggest closing this

Harsh4902:
Thanks @Caesarsage, I'm clossing this PR as it's already tackled.

PR #280, opened 2026-05-04
refactor: adopt cobra's error handling and properly validate inputs

mynk8:
### Description

The codebase utilizes cobra but does not use it's error handling features like cobra's PreRunE, Args, and RunE hooks were never adopted, and defer statements were skipped on error paths and never surfaced due to os.Exit(0) (cannot repro this just a speculation).

This refactor improves CI reliability by routing command validation through Cobra's standard Args / PreRunE / RunE flow. Invalid CLI usage now consistently returns a non-zero exit code while preserving actionable usage output, which is important for GitHub Actions and other automated environments that rely on process status.

The changes fix the problems with input validation with proper integration of cobra in the codebase.

I have added small tests for sanity checks as well, would like some feedback on this PR :)

Changes:
- All commands: Run converted to RunE correctly returning error
- test: Args ExactArgs(3), PreRunE for runner and --waitFor validation
- import: Args MinimumNArgs(1)
- import-url: Args MinimumNArgs(1)
- import-dir: Args ExactArgs(1)
- login: Args ExactArgs(1), PreRunE for --sso-port range validation
- logout: Args ExactArgs(1)
- context: conditional arg validation in RunE
- start: PreRunE for --port range validation
- Root command: add SilenceErrors alongside existing SilenceUsage
- 24 os.Exit(1) calls removed, replaced by error returns
- log.Fatalf calls in start/stop replaced by fmt.Errorf returns
- Port validation tests for --port and --sso-port

## Bugs

Addresses #281

PR #344, opened 2026-05-08
refactor(cmd): replace os.Exit/log.Fatalf with RunE and add Args validators

### LINKED ISSUE #281: CLI Commands use os.Exit/log.Fatalf instead of cobra's RunE, Args, and PreRun
### Reason/Context

There are several subtle input validation bugs in the arguments and the code is not fully utilizing the idiomatic features of cobra for handling errors and lifecycle.

### Description

Here's a small list I can think of

- **No `Args` validators** -- manual `len(args)` checks with `cmd.HelpFunc()` + `os.Exit(1)` instead of cobra's `ExactArgs`, `MinimumNArgs`
- **No `PreRunE`** -- flag validation happens inside `Run` after network calls have started, instead of before
- **`SilenceErrors` not set** -- cobra prints errors on top of the command's own error output

Some example validation loopholes

- `microcks login --sso-port 0` / `--sso-port 65536` — passes invalid ports to HTTP listener
- `microcks test --waitFor 5abc` — crashes with `os.Exit(1)` (this was addressed in PR #271 as well)

### Implementation ideas

_No response_

Harishrs2006:
1) add `SilenceErrors: true` to root command to prevent errors being printed twice by both Cobra and `main()`.
2)   `import`, `import-url`, `import-dir`: `cobra.ExactArgs(1)` replaces manual `len(args)` guard and `os.Exit`; `RunE` returns wrapped errors.
3)  `test`: `cobra.ExactArgs(3)` replaces direct `os.Args` inspection; `PreRunE` validates `--waitFor` format and runner type before any network call; extract `parseWaitForMilliseconds` helper for testability
4)   `start`, `stop`: `RunE` replaces `log.Fatalf` with `fmt.Errorf`
5) all subcommands now return errors instead of calling `os.Exit` — errors propagate cleanly to `main()`.

Fixes #281

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -27,9 +27,10 @@ func NewCommad() *cobra.Command {
 	var clientOpts connectors.ClientOptions
 
 	command := &cobra.Command{
-		Use:          "microcks",
-		Short:        "A CLI tool for Microcks",
-		SilenceUsage: true,
+		Use:           "microcks",
+		Short:         "A CLI tool for Microcks",
+		SilenceUsage:  true,
+		SilenceErrors: true,
 		Run: func(cmd *cobra.Command, args []string) {
 			cmd.HelpFunc()(cmd, args)
 		},
--- a/cmd/import.go
+++ b/cmd/import.go
@@ -17,7 +17,6 @@ package cmd
 
 import (
 	"fmt"
-	"os"
 	"strconv"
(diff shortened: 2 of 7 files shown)


Harishrs2006:
Closing in favour of #280 which covers the same refactor more completely 
(also adds login/logout/context validation and --port range checks).

PR #353, opened 2026-05-09
fix(import): preserve paths containing colons when parsing primary su…

### LINKED ISSUE #352: microcks import truncates Windows absolute paths when parsing :primary suffix
### Describe the bug

`microcks import` supports an optional `:true` / `:false` suffix to mark whether an imported artifact is the main artifact. However, the current parser at `cmd/import.go:126` uses `strings.Split(f, ":")`, which also splits Windows-absolute paths on the drive-letter colon. A path such as:

```text
C:\Temp\api.yaml
```

is therefore truncated to just `C`, and the remainder is treated as the boolean suffix — which then fails to parse, falls back to `mainArtifact = true`, and the upload step opens `C` (which does not exist).

---

### Expected behavior

Windows-absolute paths should be preserved as full file paths.

`C:\Temp\api.yaml` should be parsed as:

```text
path         = C:\Temp\api.yaml
mainArtifact = true
```

`C:\Temp\api.yaml:false` should be parsed as:

```text
path         = C:\Temp\api.yaml
mainArtifact = false
```

The existing documented suffix behavior should continue to work for relative paths:

```text
./api.yaml         -> path = ./api.yaml, mainArtifact = true
./api.yaml:true    -> path = ./api.yaml, mainArtifact = true
./api.yaml:false   -> path = ./api.yaml, mainArtifact = false
```

In short: the parser should be able to tell the *drive-letter* colon apart from the *suffix-separator* colon.

---

### Actual behavior

The current parser splits on every colon and only ever inspects `parts[1]`.

A Windows path like `C:\Temp\api.yaml` is parsed as:

```text
parts        = ["C", "\Temp\api.yaml"]
path         = "C"
mainArtifact = true   (strconv.ParseBool fails on "\Temp\api.yaml", default is true)
```

A Windows path with an explicit suffix like `C:\Temp\api.yaml:false` is parsed as:

```text
parts        = ["C", "\Temp\api.yaml", "false"]
path         = "C"
mainArtifact = true   (only parts[1] is read; the trailing "false" is never seen)
```

In both cases the subsequent upload fails with `open C: no such file or directory`, with no hint that the parser silently truncated the path.

---

### How to Reproduce?

1. On Windows, place an OpenAPI file at `C:\Temp\api.yaml`.
2. Build the CLI: `make build-local`.
3. Run:
   ```bash
   ./build/dist/microcks.exe import C:\Temp\api.yaml --microcksURL http://localhost:8585
   ```
4. Observe the failure: `open C: no such file or directory` (sometimes preceded by `Cannot parse '\Temp\api.yaml' as Bool, default to true`).

The same shape of bug is reproducible on Linux/macOS by passing a path that contains a colon, e.g. `path:with:colon.yaml`.

---

_No response_

---

_No response_

---

### Additional information

**Root cause.** `cmd/import.go:125-131` parses each comma-separated entry by:

```go
if strings.Contains(f, ":") {
    pathAndMainArtifact := strings.Split(f, ":")
    f = pathAndMainArtifact[0]
    mainArtifact, err = strconv.ParseBool(pathAndMainArtifact[1])
    ...
}
```

hazelmayank:
## Summary

`microcks import` previously parsed the optional `:true` / `:false` suffix with `strings.Split(input, ":")[0]`, which splits on the **first** colon. Any input containing more than one colon — including Windows-absolute paths such as `C:\Temp\api.yaml` — had its path truncated to the part before the first colon, and the rest was passed to `strconv.ParseBool`. The eventual upload failed with `open C: no such file or directory`, with no hint that the parser had silently eaten the path.

This PR fixes the parser to detect the boolean suffix from the **right** using `strings.LastIndex` + `strconv.ParseBool`, so paths containing colons (Windows-absolute paths, paths with multiple colons) are preserved while the documented `:true` / `:false` suffix continues to work exactly as before.

---

## Why

- Windows users are blocked from using absolute paths with `microcks import` today.
- The failure mode (`open C: no such file or directory`) gives no hint that the parser truncated the path.
- All currently-documented inputs (`./api.yaml`, `./api.yaml:true`, `./api.yaml:false`, `./api.yaml:1`, …) keep working identically after the fix — this is a strict superset of previous behavior.

---

## What changed

- Extracted a small pure helper `parseImportEntry(input string) (string, bool)` in `cmd/import.go`.
- The helper inspects the substring after the **last** `:` and only treats it as `mainArtifact` when `strconv.ParseBool` succeeds; otherwise the entire input is treated as the path with `mainArtifact = true`.
- Replaced the inline `if strings.Contains(f, ":") { ... }` block in the import loop with a single call to the helper.
- Removed the now-redundant `Cannot parse '...' as Bool` warning, because the new logic treats a non-bool trailing segment as part of the path rather than a parse failure.
- Added a new test file `cmd/import_test.go` with table-driven cases for the parser.

---

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l cmd/import.go cmd/import_test.go` — no output
- [x] `go test ./cmd/... -run TestParseImportEntry -v` — all table rows pass
- [x] `go test ./...` — full sweep passes
- [x] `go test ./cmd/... -count=3` — passes (no state leakage)
- [x] Manually verified on Windows: `microcks import C:\Temp\api.yaml` no longer fails with `open C: no such file or directory`
- [x] Manually verified that `./api.yaml`, `./api.yaml:true`, `./api.yaml:false`, `./api.yaml:1` all parse as before

---

## Test cases covered

The table-driven test exercises:

| Input | Path | mainArtifact |
| --- | --- | --- |
| `./api.yaml` | `./api.yaml` | `true` |
| `./api.yaml:true` | `./api.yaml` | `true` |
| `./api.yaml:false` | `./api.yaml` | `false` |
| `./api.yaml:TRUE` | `./api.yaml` | `true` |
| `./api.yaml:0` | `./api.yaml` | `false` |
| `C:\Temp\api.yaml` | `C:\Temp\api.yaml` | `true` |
| `C:\Temp\api.yaml:false` | `C:\Temp\api.yaml` | `false` |
| `C:\Temp\api.yaml:true` | `C:\Temp\api.yaml` | `true` |
| `path:with:colon.yaml` | `path:with:colon.yaml` | `true` |
| `path:with:colon.yaml:false` | `path:with:colon.yaml` | `false` |
| `./api.yaml:notabool` | `./api.yaml:notabool` | `true` |

---

## Files changed

- `cmd/import.go` — parser refactored; `strconv` import retained for `ParseBool`.
- `cmd/import_test.go` — new file; first test in this package's import-command path.

---

## Notes

- No flag changes; no behavior change for currently-working inputs.
- A similar bug exists in `cmd/importURL.go` for URLs that include a port (e.g. `https://localhost:8080/api.yaml`). It is intentionally left out of this PR to keep scope tight; happy to send a follow-up if maintainers prefer that shape.
- This is one of my early contributions to the repo; I am still learning Go and welcome any feedback on idiom or structure.

Closes #352

### DIFF
--- a/cmd/import.go
+++ b/cmd/import.go
@@ -118,18 +118,10 @@ func NewImportCommand(globalClientOpts *connectors.ClientOptions) *cobra.Command
 			// Handle multiple specification files separated by comma.
 			sepSpecificationFiles := strings.Split(specificationFiles, ",")
 			for _, f := range sepSpecificationFiles {
-				mainArtifact := true
-				var err error
-
-				// Check if mainArtifact flag is provided.
-				if strings.Contains(f, ":") {
-					pathAndMainArtifact := strings.Split(f, ":")
-					f = pathAndMainArtifact[0]
-					mainArtifact, err = strconv.ParseBool(pathAndMainArtifact[1])
-					if err != nil {
-						fmt.Printf("Cannot parse '%s' as Bool, default to true\n", pathAndMainArtifact[1])
-					}
-				}
+				// Parse optional :true / :false suffix. Preserves paths
+				// containing colons (e.g. Windows C:\Temp\api.yaml).
+				path, mainArtifact := parseImportEntry(f)
+				f = path
 
 				// Try uploading this artifact.
 				msg, err := mc.UploadArtifact(f, mainArtifact)
@@ -188,4 +180,26 @@ func NewImportCommand(globalClientOpts *connectors.ClientOptions) *cobra.Command
 
 	importCmd.Flags().BoolVar(&watch, "watch", false, "Keep watch on file changes and re-import it on change")
 	return importCmd
+}
+
+// parseImportEntry parses a single comma-separated entry of the import
+// command argument. The accepted forms are "<path>" and
+// "<path>:<bool>", where <bool> is anything strconv.ParseBool accepts
+// (true, false, TRUE, FALSE, 1, 0, t, f, ...). Because <path> may
+// itself contain colons (e.g. Windows-absolute paths such as
+// C:\Temp\api.yaml), the optional boolean suffix is detected from the
+// right: only the substring after the last ':' is considered, and
+// only when it parses as a bool is it treated as the mainArtifact
+// flag. Otherwise the whole input is returned unchanged with
+// mainArtifact defaulting to true.
+func parseImportEntry(input string) (string, bool) {
+	idx := strings.LastIndex(input, ":")
+	if idx < 0 {
+		return input, true
+	}
+	b, err := strconv.ParseBool(input[idx+1:])
+	if err != nil {
+		return input, true
+	}
+	return input[:idx], b
 }
--- a/cmd/import_test.go
+++ b/cmd/import_test.go
@@ -1,0 +1,50 @@
+/*
+ * Copyright The Microcks Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package cmd
+
+import (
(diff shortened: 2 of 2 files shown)


Harishrs2006:
the `LastIndex` + `ParseBool` approach is the right way to handle this.
 
 `cmd/importURL.go` has the same colon-splitting bug and isn't touched by this PR. Also, PR #291 is already open and covers both `import.go` and `importURL.go` together

hazelmayank:
Oh I can see a broader and similar issue is being addressed already

PR #360, opened 2026-05-10
fix(rand): return errors for invalid random string inputs

### LINKED ISSUE #359: pkg/util/rand helpers panic on invalid input instead of returning errors
### Describe the bug

`pkg/util/rand` provides two helpers — `String` and `StringFromCharset` — that both return `(string, error)`. The signature implies invalid inputs should surface as a Go `error`, but some invalid inputs currently cause the process to panic instead:

- `StringFromCharset(1, "")` panics in `crypto/rand.Int` because the upper bound is `<= 0`.
- `String(-1)` panics in `make([]byte, -1)` because the slice length is negative.

Since both functions already declare an `error` return, callers naturally expect that path to be used. Today they cannot recover — the helpers crash the host process before any caller code can react.

---

### Expected behavior

Invalid inputs should return a clear error instead of panicking:

```text
StringFromCharset(1, "")  ->  "", error("rand: cannot generate ... from empty charset")
String(-1)                ->  "", error("rand: requested length -1 is negative")
```

`n == 0` should keep returning `("", nil)` regardless of the charset, since that's the natural no-op result for both helpers.

---

### Actual behavior

`StringFromCharset(1, "")` panics with:

```text
panic: crypto/rand: argument to Int is <= 0
```

`String(-1)` panics with:

```text
panic: runtime error: makeslice: len out of range
```

In both cases the process crashes; the `error` return value is never set.

---

### How to Reproduce?

Reproduced locally on the latest `master`.

**Reproducer for empty charset:**

```go
package main

import (
    "fmt"

    r "github.com/microcks/microcks-cli/pkg/util/rand"
)

func main() {
    fmt.Println("Testing empty charset...")
    value, err := r.StringFromCharset(1, "")
    fmt.Printf("value=%q err=%v\n", value, err)
}
```

Output:

```text
Testing empty charset...
panic: crypto/rand: argument to Int is <= 0
```

**Reproducer for negative length:**

```go
package main

import (
    "fmt"

    r "github.com/microcks/microcks-cli/pkg/util/rand"
)

func main() {
    fmt.Println("Testing negative length...")
    value, err := r.String(-1)
    fmt.Printf("value=%q err=%v\n", value, err)
}
```

Output:

```text
Testing negative length...
panic: runtime error: makeslice: len out of range
```

In both cases the program exits via panic before the `err=` line can print.

---

_No response_

---

_No response_

---

### Additional information

**Root cause.** `StringFromCharset` in `pkg/util/rand/rand.go` does not validate `n` or `charset` before allocating the output slice and calling `crypto/rand.Int`:

```go
func StringFromCharset(n int, charset string) (string, error) {
    b := make([]byte, n)
    maxIdx := big.NewInt(int64(len(charset)))
    for i := 0; i < n; i++ {
        randIdx, err := rand.Int(rand.Reader, maxIdx)
        ...
    }
    ...
}
```

- `n < 0`                       → `make([]byte, n)` panics with `makeslice: len out of range`.
- `n > 0` with `charset == ""`  → `rand.Int(rand.Reader, big.NewInt(0))` panics with `argument to Int is <= 0`.
- `n == 0`                      → already works correctly (loop doesn't execute, returns `("", nil)`).

`String` calls `StringFromCharset` and inherits the negative-length panic.

**Proposed fix.** Add two guard clauses at the top of `StringFromCharset`:

```go
if n < 0 {
    return "", fmt.Errorf("rand: requested length %d is negative", n)
}
if n > 0 && len(charset) == 0 {
    return "", fmt.Errorf("rand: cannot generate %d-character string from empty charset", n)
}
```

This preserves every currently-working input (including `n == 0`) and turns both panic paths into descriptive errors. The public signatures stay the same, so no callers (including the SSO PKCE flow in `cmd/login.go`) need to change.

A small test file `pkg/util/rand/rand_test.go` would lock this in with table-driven cases for both valid and invalid inputs — the package currently has no test coverage.

I would like to send a small PR for this if the maintainers agree on the approach.

hazelmayank:
## Summary

This PR fixes two panic cases in `pkg/util/rand`. The helpers `String` and `StringFromCharset` both declare `(string, error)` returns, but two invalid inputs caused the process to panic instead of returning an error:

- `StringFromCharset(1, "")` panicked because `crypto/rand.Int` was called with `max = 0`.
- `String(-1)` panicked because `make([]byte, -1)` rejected the negative length.

This PR adds input validation before allocation and random-index generation, so invalid inputs now return descriptive errors instead of crashing the host process.

---

## Why

- Both functions already declare an `error` return; the panic paths were unreachable for callers who expected to handle them as errors.
- The SSO PKCE flow in `cmd/login.go` is the only current caller and is unaffected, because it always passes positive `n` and a non-empty charset.
- Adds the first unit tests to a package that previously had no test coverage.

---

## What changed

- Added two guard clauses at the top of `StringFromCharset`:
  - `n < 0` returns `"rand: requested length %d is negative"`.
  - `n > 0` with `len(charset) == 0` returns `"rand: cannot generate %d-character string from empty charset"`.
- `n == 0` continues to return `("", nil)` regardless of charset, preserving the existing no-op behavior.
- Updated the doc comment on `StringFromCharset` to spell out the new error contract.
- Added `pkg/util/rand/rand_test.go` with table-driven cases for both valid and invalid inputs. No statistical or distribution assertions — only deterministic length and charset-membership checks.

---

## Behavior preserved for valid inputs

```text
String(0)                     ->  "", nil
String(n)                     ->  string of length n, nil
StringFromCharset(0, "")      ->  "", nil
StringFromCharset(0, "abc")   ->  "", nil
StringFromCharset(10, "abc")  ->  10-character string using only a/b/c, nil
```

Public signatures are unchanged.

---

## Test plan

- [x] `gofmt -l pkg/util/rand/rand.go pkg/util/rand/rand_test.go` — no output
- [x] `git diff --check -- pkg/util/rand/rand.go pkg/util/rand/rand_test.go` — clean
- [x] `go test -v ./pkg/util/rand/...` — all cases pass
- [x] `go vet ./...` — clean
- [x] `go test ./...` — full sweep passes
- [x] `make build-local` — builds successfully

---

## Files changed

- `pkg/util/rand/rand.go` — two guard clauses + expanded doc comment.
- `pkg/util/rand/rand_test.go` — new file; first tests in this package.

---

## Notes

- I did not include `go test -cover` because my local Go toolchain reports `go: no such tool "covdata"`. The PR is scoped only to `pkg/util/rand` and its tests, so the missing coverage report does not affect the change itself; CI will produce a coverage report on its own toolchain if the project requires one.
- No production-code callers needed to change.
- No new dependencies.

Fixes #359

### DIFF
--- a/pkg/util/rand/rand.go
+++ b/pkg/util/rand/rand.go
@@ -14,7 +14,16 @@ func String(n int) (string, error) {
 }
 
 // StringFromCharset generates, from a given charset, a cryptographically-secure pseudo-random string of a given length.
+//
+// Returns an error when n is negative, or when n > 0 and charset is empty.
+// n == 0 returns "" with no error regardless of the charset.
 func StringFromCharset(n int, charset string) (string, error) {
+	if n < 0 {
+		return "", fmt.Errorf("rand: requested length %d is negative", n)
+	}
+	if n > 0 && len(charset) == 0 {
+		return "", fmt.Errorf("rand: cannot generate %d-character string from empty charset", n)
+	}
 	b := make([]byte, n)
 	maxIdx := big.NewInt(int64(len(charset)))
 	for i := 0; i < n; i++ {
--- a/pkg/util/rand/rand_test.go
+++ b/pkg/util/rand/rand_test.go
@@ -1,0 +1,83 @@
+package rand
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestString_ValidLengths(t *testing.T) {
+	tests := []struct {
+		name string
+		n    int
+	}{
+		{"zero length", 0},
+		{"single char", 1},
+		{"medium length", 24},
+		{"PKCE-style length", 43},
+		{"long", 128},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := String(tt.n)
+			require.NoError(t, err)
+			assert.Len(t, got, tt.n)
+			for i := 0; i < len(got); i++ {
+				assert.True(t, strings.IndexByte(letterBytes, got[i]) >= 0,
+					"byte %q at position %d not in default letter charset", got[i], i)
+			}
+		})
+	}
+}
+
+func TestString_NegativeLengthReturnsError(t *testing.T) {
+	got, err := String(-1)
+	require.Error(t, err)
+	assert.Equal(t, "", got)
+}
+
+func TestStringFromCharset_ValidInputs(t *testing.T) {
+	tests := []struct {
+		name    string
+		n       int
+		charset string
+	}{
+		{"empty length, empty charset", 0, ""},
+		{"empty length, non-empty charset", 0, "abc"},
+		{"single byte from single-byte charset", 1, "x"},
+		{"PKCE verifier charset", 43, "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-._~"},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := StringFromCharset(tt.n, tt.charset)
+			require.NoError(t, err)
+			assert.Len(t, got, tt.n)
+			for i := 0; i < len(got); i++ {
+				assert.True(t, strings.IndexByte(tt.charset, got[i]) >= 0,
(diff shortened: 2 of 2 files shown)


hazelmayank:
@Harsh4902 Could you please review this pr?

Caesarsage:
Thanks for this — you spotted and fixed both panics first (#359). #494 has since received an LGTM and covers the same two cases with tests, so this will likely close as a duplicate when it merges. Appreciate the original report and fix; hope to see you on the other open items.

PR #363, opened 2026-05-11
Remove redundant Running status check

Syedowais312:
#### Summary

Removes a redundant `instance.Status == "Running"` check before the switch statement in the start command.

The `"Running"` case was already handled inside the switch, so the extra if-condition was unnecessary.

#### Changes

- Removed duplicate Running status check
- Kept existing behavior unchanged

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -46,12 +46,6 @@ microcks start --name [name of you container/instance]`,
 			if instance == nil {
 				instance = &config.Instance{}
 			}
-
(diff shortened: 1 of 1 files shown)


mynk8:
There is an already open PR #290

Syedowais312:
> There is an already open PR #290

Oops, I didn’t see PR #290.
Thanks for letting me know.

Syedowais312:
#290

PR #372, opened 2026-05-12
fix: stream multipart upload in UploadArtifact to avoid unbounded mem…

### LINKED ISSUE #324: Unbounded memory allocation in UploadArtifact due to in-memory multipart buffering
### Describe the bug

While importing API specifications, the CLI loads the entire target file into memory before executing the HTTP POST request. In pkg/connectors/microcks_client.go, the UploadArtifact method relies on a &bytes.Buffer{} backed by io.Copy() to construct the multipart form data.

Because bytes.Buffer buffers data into RAM, uploading large enterprise files (e.g., massive Postman Collections or deeply nested OpenAPI schemas) forces Go to allocate heap memory equivalent to the total size of the uploaded file. This causes severe memory spikes during the execution of microcks import or microcks import-dir.

### Expected behavior

The CLI should maintain a near-zero memory footprint regardless of the artifact size. The multipart upload should stream the payload directly from disk to the network socket.

### Actual behavior

The CLI eagerly loads the entire API specification file into the system's RAM before the upload begins. On constrained CI/CD runners (e.g., containers with 512MB RAM limits), running microcks import-dir across a repository containing large collections triggers Out-Of-Memory (OOM) kills, causing silent pipeline failures.

### How to Reproduce?

Reproducer:
- Create a massive dummy text file locally to simulate a large API spec (e.g., a 200MB file): dd if=/dev/urandom of=large_spec.json bs=1M count=200
- Run the CLI import command on this file: microcks import large_spec.json --microcksURL=http://localhost:8080/api/
- While the command is running, monitor the process footprint via top or htop.
- Observe the CLI's memory consumption spike linearly by ~200MB+ as it buffers the entire file into the heap before sending.

master (and v1.0.3)   

CLI Binary

### Additional information

_No response_

abhaygoudannavar:
### Description
- Replaced in-memory `bytes.Buffer` with `io.Pipe()` streaming in `UploadArtifact` (`pkg/connectors/microcks_client.go`). Multipart form data is now written in a background goroutine and streamed directly from disk to the HTTP request, keeping memory usage constant regardless of artifact file size.
- Replaced two `panic()` calls with proper `fmt.Errorf` error returns in the upload path, matching the style of recent fixes (e.g., `f9f282c`).
- Added `TestUploadArtifactStreamsWithoutBuffering` test (`pkg/connectors/microcks_client_test.go`) covering file content integrity, filename, `mainArtifact` field, and HTTP response handling.

**Test results:**
$ go test ./pkg/connectors/... ok github.com/microcks/microcks-cli/pkg/connectors 0.022s

### Related issue(s)
Fixes #324

### DIFF
--- a/pkg/connectors/microcks_client.go
+++ b/pkg/connectors/microcks_client.go
@@ -425,31 +425,41 @@ func (c *microcksClient) UploadArtifact(specificationFilePath string, mainArtifa
 	}
 	defer file.Close()
 
-	// Create a multipart request body, reading the file.
-	body := &bytes.Buffer{}
-	writer := multipart.NewWriter(body)
-	part, err := writer.CreateFormFile("file", filepath.Base(specificationFilePath))
-	if err != nil {
-		return "", err
-	}
-	_, err = io.Copy(part, file)
-	if err != nil {
-		panic(err.Error())
-	}
+	// Use io.Pipe to stream the multipart form data directly to the HTTP
+	// request without buffering the entire file in memory.
+	pr, pw := io.Pipe()
+	writer := multipart.NewWriter(pw)
 
-	// Add the mainArtifact flag to request.
-	_ = writer.WriteField("mainArtifact", strconv.FormatBool(mainArtifact))
+	// Write the multipart form data in a background goroutine so the pipe
+	// reader can be consumed concurrently by the HTTP request.
+	errCh := make(chan error, 1)
+	go func() {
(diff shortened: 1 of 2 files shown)


abhaygoudannavar:
@Harsh4902 i worked on #324 any changes is needed in the pr?

Harsh4902 (approved):
Thanks @abhaygoudannavar for this contribution. LGTM

Issue #375, opened 2026-05-12
Bug : DownloadArtifact reading request body instead of response body


Manar-Elhabbal7:
### Describe the bug

The `DownloadArtifact` function incorrectly reads from **req.Body** instead of **resp.Body** after executing the HTTP request.

## Current implementation:
`respBody, err := io.ReadAll(req.Body)`

This causes the function to read the request payload rather than the server response body, which may lead to empty or invalid artifact content when using import-url functionality.

## Expected behavior:
Microcks has discovered 'SUCCESS_FROM_SERVER'

## Actual  behavior:
Microcks has discovered ''

Caesarsage:
Thank you @Manar-Elhabbal7 

This has been fixed on master, DownloadArtifact now reads resp.Body, and TestDownloadArtifactReturnsResponseBody guards the regression. 

I believe this can be closed.

PR #376, opened 2026-05-12
fix: read from response body in DownloadArtifact

Manar-Elhabbal7:
### Description

When using import-url, the CLI was showing an empty message:
Microcks has discovered ''.

This happened because the code was reading the response from the wrong place (req.Body instead of resp.Body), so it couldn’t get the server message

**before solving it :**

image

**after solving it**

Screenshot from 2026-05-12 23-26-10

### Related issue(s) #375


Caesarsage:
As per my comment here https://github.com/microcks/microcks-cli/issues/375#issuecomment-5309179021

I believe this can be closed

Issue #367, opened 2026-05-11
feat: add --verbose flag to import command for detailed upload feedback


A-d-i-t-y:
### Reason/Context

Why we need this improvement:
The import command currently gives no feedback during execution. 
Users cannot tell if the upload is in progress or why it silently failed.

How this will help:
A --verbose flag will print artifact path, server response status, 
and error details — making it much easier to debug failed imports 
in CI/CD pipelines without extra tooling.

Motivation:
This aligns with the LFX Mentorship goal of improving the developer 
experience for Microcks CLI users.

## Problem
The `microcks import` command gives no detailed output during execution. 
Users cannot tell whether the upload is in progress, what the server 
responded, or why an import silently failed.

## Proposed Solution
Add a `--verbose` flag to the `import` command that prints:
- The artifact path being uploaded
- The server response status
- Any error details from the server response body

## Why This Matters
This improves the developer experience for CI/CD pipelines where 
silent failures are hard to debug without re-running with extra tooling.

### Implementation ideas

- cmd/import.go: add --verbose flag definition
- pkg/connectors/microcks_client.go: log request/response details 
  when verbose mode is enabled
- documentation/cmd/import.md: document the new flag with example

A-d-i-t-y:
Hi @Harsh4902, I am Aditya, a LFX Mentorship applicant for the 
IDE Integration and Local Dev Loop Enhancement project. I would 
love to work on this issue. Could you please assign it to me?

Harsh4902:
@A-d-i-t-y Just need one clearification, I think we already have a persistant `--verbose` flag at global level which is applied to all the commands.

A-d-i-t-y:
Hi @Harsh4902, thank you for the clarification! You are absolutely right, I missed the global --verbose flag. 

In that case, could you suggest what would be a more useful improvement for the import command? I would love to work on something that adds real value. Happy to take any direction from you!

Harsh4902:
Clossing this as there is no such issue.

PR #383, opened 2026-05-14
feat: add --verbose flag to import command

A-d-i-t-y:
## Summary
Adds `--verbose` flag to the `import` command for detailed upload feedback.

## Problem
The `microcks import` command gives no detailed output during execution. 
Users cannot tell whether the upload is in progress or why an import 
silently failed in CI/CD pipelines.

## Solution
- `--verbose` flag prints artifact path before uploading
- Prints server response after successful import
- No breaking changes to existing behavior

## Example output

### DIFF
--- a/cmd/import.go
+++ b/cmd/import.go
@@ -30,6 +30,7 @@ import (
 
 func NewImportCommand(globalClientOpts *connectors.ClientOptions) *cobra.Command {
 	var watch bool
+	var verbose bool
 
 	var importCmd = &cobra.Command{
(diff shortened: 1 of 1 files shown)


A-d-i-t-y:
Hi @Harsh4902 @Caesarsage, I have raised this PR to add --verbose flag to the import command as discussed in issue #367. This is a minimal change with no breaking changes to existing behavior. Would love your feedback!

Harsh4902:
Closing this PR as it's not required.

PR #384, opened 2026-05-14
fix: return non-zero on import-dir failures

### LINKED ISSUE #379: bug: import-dir exits 0 after partial artifact import failures
### Describe the bug

`microcks import-dir` can report failed artifact imports but still complete with exit code `0`.

The command already tracks failed imports through `ImportResult.FailedCount`, but after printing the final import summary it does not return a non-zero exit code when `FailedCount > 0`.

This is problematic for scripts and CI pipelines because a partial import failure is treated as a successful command execution.

### Expected behavior

When one or more files fail to import, `microcks import-dir` should:

- continue processing the remaining files
- print the import summary
- exit with a non-zero status code, for example `1`

If all files are imported successfully, it should continue exiting with status code `0`.

### Actual behavior

When `ImportDirectory` completes with partial failures, the command prints the failed files and final summary, but then returns normally.

Because the Cobra command returns without calling `os.Exit(1)` or returning an error, the process exits with status code `0`.

Image

### How to Reproduce?

This can be observed from the current command flow in `cmd/importDir.go`.

Reproducer:

1. Make `ImportDirectory` process multiple specification files.
2. Have one `UploadArtifact` call fail while another succeeds.
3. `ImportDirectory` returns a result with `FailedCount > 0` and `err == nil`.
4. `NewImportDirCommand` prints the summary but does not check `result.FailedCount`.
5. The command returns normally, so the process exits with status code `0`.

The existing unit test `TestImportDirectory/partial_failure` already demonstrates the partial-failure case at the business-logic level: one file succeeds, one file fails, and `FailedCount` is set to `1`.

Resources to reproduce the behaviour:
1. `cmd/importDir.go`
2. `cmd/importDir_test.go`
3. Existing test case: `TestImportDirectory/partial_failure`

Microcks-CLI  1.0.2

_No response_

### Additional information

A possible fix is to check `result.FailedCount` after printing the summary:

```go
if result.FailedCount > 0 {
    os.Exit(1)
}

cotishq:
### Description

- Converts `import-dir` command execution to Cobra `RunE` so validation/config/import errors are returned instead of silently exiting successfully.
- Returns a non-zero exit status when one or more artifact imports fail, including partial and total import failure cases.
- Updates the failure summary message to make failed imports explicit in CLI/CI logs.
- Adds tests for import result failure handling and success/failure summary output.

### Related issue(s)

Fixes #379 

### Proof of Work
Screenshot From 2026-05-14 21-09-19

### DIFF
--- a/cmd/importDir.go
+++ b/cmd/importDir.go
@@ -87,12 +87,21 @@ type ValidationError struct {
 	Message string
 }
 
+type ImportFailuresError struct {
+	FailedCount int
+	TotalFiles  int
+}
+
 func (e ImportError) Error() string {
 	return fmt.Sprintf("failed to import %s: %v", e.File, e.Err)
 }
 
 func (e ValidationError) Error() string {
 	return e.Message
+}
+
+func (e ImportFailuresError) Error() string {
+	return fmt.Sprintf("%d/%d files failed to import", e.FailedCount, e.TotalFiles)
 }
 
 func NewImportDirCommand(globalClientOpts *connectors.ClientOptions) *cobra.Command {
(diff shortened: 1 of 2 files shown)


cotishq:
@Harsh4902 would love your opinion on this

Caesarsage:
Thanks @cotishq 

This has been dixed on master (same report as #473), partial failures now produce a non-zero exit via importDirectoryPartialFailure.

 I believe this can be closed.

cotishq:
@Caesarsage ok, i will close this

Issue #388, opened 2026-05-15
Cleanup: CreateTestResult payload construction and minor typo in cmd


PranjaliBhardwaj:
I found the code in pkg/connectors/microcks_client.go and see that the CreateTestResult function is creating the JSON body using string concatenation. A little brittle and difficult to add more fields to the test request later on.

Also, I noticed that in cmd/cmd.go, the main entry point is called NewCommad (and not NewCommand).

I'm happy to put together a PR to:

Utilise a correct struct and json.Marshal test request payload.
Correct the typo in the NewCommad package and main.go in the CLI.

Caesarsage:
Both parts are done on master: CreateTestResult now uses a typed struct + json.Marshal, and the entrypoint was renamed to cmd.NewCommand() during the error-handling refactor (#490). 

I believe this can be closed.

Caesarsage:
And the underlying PR has also be merged

PR #389, opened 2026-05-15
refactor: use struct for test request and fix CLI typo

PranjaliBhardwaj:
### Description
Looking through pkg/connectors and it still appeared that CreateTestResult was manually creating the test result using string concatenation to build the JSON. It was a bit of a chore to read and I'm not sure why we needed to add more fields later on, but I replaced it with a proper testRequest struct and json.Marshal.

I used json.We don't need to convert those optional "flags" (like filteredOperations) to JSON because they are already JSON strings from the CLI. Don't worry, I'm not about to unmarshal them and then put them back in the payload, but this way we are able to use omitempty properly.

Also found a typo in the main entry point: NewCommad -> NewCommand. I can't tell if it was by design or not, but I could not find it in any other place than main.go, so I just fixed it.

### Related issue(s) : #388 

Implementation Summary
Introduced a new struct called internal testRequest into microcks_client.go.
Replaced manual concatenation of strings in CreateTestResult with json.Marshal.
Changed the name of NewCommad to NewCommand in file cmd/cmd.go and main.go.
How I tested this
Rebuilt locally to ensure that the rename didn't cause any problems.
Tested JSON output for omitempty, which seems to work for secretName.
No new test case created for this because it is largely a refactoring of existing behavior, but the existing client tests appear happy.

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -22,7 +22,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
-func NewCommad() *cobra.Command {
+func NewCommand() *cobra.Command {
 
 	var clientOpts connectors.ClientOptions
 
--- a/main.go
+++ b/main.go
@@ -8,7 +8,7 @@ import (
 )
 
 func main() {
-	command := cmd.NewCommad()
+	command := cmd.NewCommand()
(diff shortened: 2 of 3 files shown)


Harsh4902 (approved):
LGTM

PR #391, opened 2026-05-15
fix(test): replace os.Args guard with cobra.ExactArgs(3) to prevent panic

### LINKED ISSUE #390: Bug: `microcks test` panics with index out of range when global flags are used and <runner> arg is omitted
### Summary

The `test` command guards positional argument count using `len(os.Args) < 4` (line 50 of `cmd/test.go`). When the command is invoked with global persistent flags  -- the pattern shown in the project README for CI usage -- `os.Args` contains enough elements to pass the guard, but Cobra's parsed `args` slice (which has all flags stripped) may have fewer than 3 elements. Accessing `args[2]` then causes a **runtime panic** with a full stack trace instead of a clean error message.

### Steps to Reproduce

Build the binary and run with global flags but omit the `<runner>` argument:

```bash
microcks test \
  --microcksURL=http://localhost:8080 \
  --keycloakClientId=foo \
  --keycloakClientSecret=bar \
  'MyAPI:1.0' http://localhost:3000
```

### Actual Behavior

```
goroutine 1 [running]:
runtime error: index out of range [2] with length 2
...full stack trace...
```

### Expected Behavior

```
Error: accepts 3 arg(s), received 2
Usage:
  microcks test <apiName:apiVersion> <testEndpoint> <runner> [flags]
```

### Root Cause

The guard was written before global persistent flags (`--microcksURL`, `--keycloakClientId`, `--keycloakClientSecret`, etc.) were added to the root command. At that point `os.Args` and the positional `args` slice were in sync. After global flags were introduced, `os.Args` inflates with each flag, but `args` (as populated by Cobra after flag parsing) remains positional-only.

```go
// BEFORE (broken): os.Args includes global flags, so the guard is bypassed
if len(os.Args) < 4 {
    ...
}
serviceRef   := args[0]
testEndpoint := args[1]
runnerType   := args[2]  // panics when args has only 2 elements
```

### Fix

Use `cobra.ExactArgs(3)` on the command definition -- the idiomatic Cobra approach -- which validates argument count before `Run` is ever called and prints the usage hint automatically.

### Additional Notes

- The three individual `strings.HasPrefix(x, "-")` guards that follow the count check are also dead code: Cobra strips all flags before handing `args` to `Run`, so none of those args can ever start with `-`.
- - The runner validation error message listed `SOAP` (not a valid runner) instead of `SOAP_HTTP`.
- - - A regression test is included in the fix PR to assert exit code 1 and no panic stack trace.
### Environment

- `microcks-cli` version: latest `master`
- - Go version: 1.22+
- - - Reproduced when global flags are passed (typical CI invocation per README)
- - -

gyanranjanpanda:
## Summary

Fixes #390

The `test` command was guarding positional argument count with `len(os.Args) < 4`. This check includes global persistent flags (`--microcksURL`, `--keycloakClientId`, `--keycloakClientSecret`) in the count. When the command is invoked with global flags but the `<runner>` argument is omitted - the standard CI invocation pattern shown in the README - `os.Args` has enough elements to pass the guard, but Cobra's `args` slice (flags stripped) has only 2 elements. `args[2]` then causes a **runtime panic** with a stack trace instead of a clean error message.

## Changes

- **`cmd/test.go`**: Use `cobra.ExactArgs(3)` on the command definition (idiomatic Cobra). The framework validates arg count and prints usage before `Run` is invoked - no manual guard needed.
- - **`cmd/test.go`**: Update `Use` field to `"test <apiName:apiVersion> <testEndpoint> <runner>"` so `--help` output shows the expected signature.
- - **`cmd/test.go`**: Remove three now-dead `strings.HasPrefix(x, "-")` guards - Cobra strips all flags before passing `args` to `Run`, so none of those conditions can ever be true.
- - **`cmd/test.go`**: Remove redundant `var testResultID string` pre-declaration immediately shadowed by `:=`.
- - **`cmd/test.go`**: Fix runner validation error message: listed `SOAP` (not in `runnerChoices`) instead of `SOAP_HTTP`.
- - **`cmd/test_test.go`**: Add regression test that simulates the CI invocation pattern (global flags + missing runner) and asserts exit code 1 with no `panic:` in output.
## Before / After

**Before** (with global flags, missing `<runner>`):
```
goroutine 1 [running]:
runtime error: index out of range [2] with length 2
    cmd/test.go:57 +0x...
```

**After**:
```
Error: accepts 3 arg(s), received 2
Usage:
  microcks test <apiName:apiVersion> <testEndpoint> <runner> [flags]
```

## Testing

```bash
go test ./cmd/... -run TestTestCommandMissingRunnerWithGlobalFlagsDoesNotPanic -v
```
```
=== RUN   TestTestCommandMissingRunnerWithGlobalFlagsDoesNotPanic
--- PASS: TestTestCommandMissingRunnerWithGlobalFlagsDoesNotPanic (0.01s)
PASS
```

### DIFF
--- a/cmd/test.go
+++ b/cmd/test.go
@@ -42,35 +42,18 @@ func NewTestCommand(globalClientOpts *connectors.ClientOptions) *cobra.Command {
 	)
 	var testCmd = &cobra.Command{
 
-		Use:   "test",
+		Use:   "test <apiName:apiVersion> <testEndpoint> <runner>",
 		Short: "Run tests on Microcks",
 		Long:  `Run tests on Microcks`,
+		Args:  cobra.ExactArgs(3),
 		Run: func(cmd *cobra.Command, args []string) {
-			// Parse subcommand args first.
-			if len(os.Args) < 4 {
-				fmt.Println("test command require <apiName:apiVersion> <testEndpoint> <runner> args")
-				os.Exit(1)
-			}
 
 			serviceRef := args[0]
 			testEndpoint := args[1]
 			runnerType := args[2]
 
-			// Validate presence and values of args.
-			if len(serviceRef) == 0 || strings.HasPrefix(serviceRef, "-") {
-				fmt.Println("test command require <apiName:apiVersion> <testEndpoint> <runner> args")
-				os.Exit(1)
-			}
-			if len(testEndpoint) == 0 || strings.HasPrefix(testEndpoint, "-") {
-				fmt.Println("test command require <apiName:apiVersion> <testEndpoint> <runner> args")
-				os.Exit(1)
-			}
-			if len(runnerType) == 0 || strings.HasPrefix(runnerType, "-") {
-				fmt.Println("test command require <apiName:apiVersion> <testEndpoint> <runner> args")
-				os.Exit(1)
-			}
-			if _, validChoice := runnerChoices[runnerType]; !validChoice {
-				fmt.Println("<runner> should be one of: HTTP, SOAP, SOAP_UI, POSTMAN, OPEN_API_SCHEMA, ASYNC_API_SCHEMA, GRPC_PROTOBUF, GRAPHQL_SCHEMA")
+			if _, ok := runnerChoices[runnerType]; !ok {
+				fmt.Println("<runner> should be one of: HTTP, SOAP_HTTP, SOAP_UI, POSTMAN, OPEN_API_SCHEMA, ASYNC_API_SCHEMA, GRPC_PROTOBUF, GRAPHQL_SCHEMA")
 				os.Exit(1)
 			}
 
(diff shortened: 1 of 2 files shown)


gyanranjanpanda:
could @lbroudoux  review this pr

gyanranjanpanda:
@lbroudoux  could u review this pr

gyanranjanpanda:
hey @lbroudoux  and. @Harsh4902  i want to know about your view on this pr

Harsh4902:
@gyanranjanpanda Sorry for late response. I'm reviewing this now.

Harsh4902 (approved):
LGTM. Thanks @gyanranjanpanda for this PR.

PR #394, opened 2026-05-16
fix: continue importing remaining artifacts on failure

### LINKED ISSUE #393: fix: import command exits early on first failure when importing multiple artifacts
Why we need this improvement:
When importing multiple comma-separated files, if one file fails, 
the command immediately exits and remaining files are never imported.
Users have no visibility into which files succeeded and which failed.

How this will help:
All files should be attempted regardless of individual failures.
A summary at the end should show which files succeeded and which failed,
giving users complete feedback in one run.

Motivation:
This is a silent bug in CI/CD pipelines where partial imports can 
cause hard-to-debug failures.             

## Problem
When running:
microcks import "file1.yaml,file2.yaml,file3.yaml"

If file2.yaml fails, the command immediately calls os.Exit(1) 
and file3.yaml is never imported. The user has no idea which 
files were successfully imported and which were not.

## Proposed Fix
- Collect errors instead of exiting immediately
- Attempt all files regardless of individual failures
- Print a summary at the end showing pass/fail for each file
- Exit with code 1 only if any file failed

## Example output after fix
Importing 'file1.yaml'... success
Importing 'file2.yaml'... failed: connection refused
Importing 'file3.yaml'... success

Import completed: 2 succeeded, 1 failed

A-d-i-t-y:
## Summary
Fix early exit bug in import command when importing multiple artifacts.

## Problem
When running:
microcks import "file1.yaml,file2.yaml,file3.yaml"

If file2.yaml fails, the command immediately exits and file3.yaml 
is never imported. User has no visibility into which files succeeded.

## Fix
- Collect errors instead of exiting immediately on first failure
- Attempt all artifacts regardless of individual failures  
- Exit with code 1 only if any artifact failed

## Example output after fix
Failed to import 'file2.yaml': connection refused
Microcks has discovered 'file1.yaml'
Microcks has discovered 'file3.yaml'

## Files changed
- cmd/import.go — replaced os.Exit(1) with error collection and continue

Closes #393

### DIFF
--- a/cmd/import.go
+++ b/cmd/import.go
@@ -114,6 +114,7 @@ func NewImportCommand(globalClientOpts *connectors.ClientOptions) *cobra.Command
 					return
 				}
 			}
+			var hasErrors bool
 
 			// Handle multiple specification files separated by comma.
 			sepSpecificationFiles := strings.Split(specificationFiles, ",")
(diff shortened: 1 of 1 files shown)


A-d-i-t-y:
Hi @Harsh4902 @Caesarsage, I have raised this PR to fix the early exit bug in the import command when importing multiple artifacts as discussed in issue #393. This is a minimal change — replaced os.Exit(1) with error collection so all artifacts are attempted before exiting. Would love your feedback!

PR #396, opened 2026-05-16
fix(connectors): return error for malformed Keycloak config response

### LINKED ISSUE #395: pkg/connectors: GetKeycloakURL panics on malformed Keycloak config response
### Describe the bug

`GetKeycloakURL` in `pkg/connectors/microcks_client.go` parses the response from `/api/keycloak/config` using direct, single-value type assertions:

```go
// pkg/connectors/microcks_client.go:240-242
enabled := configResp["enabled"].(bool)
authServerURL := configResp["auth-server-url"].(string)
realmName := configResp["realm"].(string)
```

If the response is missing any of these fields, or if a field has an unexpected type, the assertion panics rather than producing the `error` value the function's signature already promises. The CLI then crashes with a Go stack trace.

---

### Expected behavior

A malformed or unexpected response from `/api/keycloak/config` should cause `GetKeycloakURL` to return a descriptive `error`. Valid responses (Keycloak enabled or Keycloak disabled) should continue to work exactly as today.

---

### Actual behavior

The CLI panics with an interface-conversion error similar to:

```text
panic: interface conversion: interface {} is nil, not bool

goroutine 1 [running]:
github.com/microcks/microcks-cli/pkg/connectors.(*microcksClient).GetKeycloakURL(...)
        pkg/connectors/microcks_client.go:240
```

The process exits via panic instead of via the `(string, error)` return path the function already declares.

---

### How to Reproduce?

Minimal Go reproducer using `httptest`:

```go
package main

import (
    "net/http"
    "net/http/httptest"

    "github.com/microcks/microcks-cli/pkg/connectors"
)

func main() {
    s := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
        w.Header().Set("Content-Type", "application/json")
        w.Write([]byte(`{}`)) // missing 'enabled', 'auth-server-url', 'realm'
    }))
    defer s.Close()

    c := connectors.NewMicrocksClient(s.URL)
    _, _ = c.GetKeycloakURL() // panics here
}
```

Run with:

```bash
go run main.go
```

Observe the panic instead of a clean error.

---

`master`

---

`go run .`

---

### Additional information

**Why this matters.** `GetKeycloakURL` sits in the connector/API layer used by every authenticated CLI command (`login`, `import`, `test`, ...). A Microcks server upgrade, a stub/mock server during local development, or a transient gateway returning HTML can crash the CLI instead of producing a clean error message. The function already returns `(string, error)`; the panic path is inconsistent with its own signature.

**Root cause.** Three single-value type assertions on a `map[string]interface{}` parsed from the response body:

```go
enabled := configResp["enabled"].(bool)
authServerURL := configResp["auth-server-url"].(string)
realmName := configResp["realm"].(string)
```

A single-value type assertion panics if the asserted type does not match. The two-value (comma-ok) form returns `(value, false)` instead, which lets the function surface the problem as an `error`.

**Proposed fix.** Switch to the comma-ok form on each assertion and return a descriptive error when a field is missing or wrong-typed. The early `return "null", nil` for the Keycloak-disabled case is preserved:

```go
enabled, ok := configResp["enabled"].(bool)
if !ok {
    return "", fmt.Errorf("Microcks /api/keycloak/config response missing or invalid 'enabled' field")
}
if !enabled {
    return "null", nil
}
authServerURL, ok := configResp["auth-server-url"].(string)
if !ok {
    return "", fmt.Errorf("Microcks /api/keycloak/config response missing or invalid 'auth-server-url' field")
}
realmName, ok := configResp["realm"].(string)
if !ok {
    return "", fmt.Errorf("Microcks /api/keycloak/config response missing or invalid 'realm' field")
}
return authServerURL + "/realms/" + realmName + "/", nil
```

This preserves both currently-working cases (enabled and disabled Keycloak) and replaces the panic with the same error-return contract the function already advertises.

**Test plan.** Extend `pkg/connectors/microcks_client_test.go` using the existing `httptest.NewServer` pattern already used by `TestDownloadArtifactReturnsResponseBody`:

1. **Valid Keycloak-enabled response** — server returns `{"enabled": true, "auth-server-url": "https://kc", "realm": "microcks"}` → expect `"https://kc/realms/microcks/"` and `nil` error.
2. **Keycloak-disabled response** — server returns `{"enabled": false}` → expect `"null"` and `nil` error.
3. **Malformed response** — server returns `{}` → expect a non-nil error and no panic.

A similar single-value-assertion pattern exists in `pkg/connectors/keycloak_client.go` (e.g. `ConnectAndGetToken`). I am intentionally calling it out here but leaving it to a separate follow-up PR to keep this issue's scope tight.

I would like to send a small PR for this if the maintainers agree on the approach.

### Expected behavior

_No response_

### Actual behavior

_No response_

### How to Reproduce?

_No response_

_No response_

_No response_

### Additional information

_No response_

hazelmayank:
## Summary

This PR makes `GetKeycloakURL` defensive when parsing the `/api/keycloak/config` response.

Previously, the function used direct type assertions for fields like `enabled`, `auth-server-url`, and `realm`. If the server returned a malformed or unexpected response, the CLI could panic with an interface conversion error instead of returning a normal error.

This PR replaces those assertions with comma-ok checks and returns descriptive errors for malformed responses.

## Why

This improves CLI reliability at the connector/API layer. Commands that communicate with a Microcks server should return clean errors instead of crashing on unexpected server responses.

## What changed

- Added safe parsing for `enabled`
- Added safe parsing for `auth-server-url`
- Added safe parsing for `realm`
- Preserved existing behavior when Keycloak is disabled
- Added tests for valid, disabled, and malformed Keycloak config responses

## Test plan

- [x] `go test -v -run TestGetKeycloakURL ./pkg/connectors/...`
- [x] `go test ./pkg/connectors/...`
- [x] `go build ./...`

Note: `go test ./...` may still show the existing `TestDeleteContext` failure on master, which is unrelated to this PR and fixed separately in #334.

Closes #395

### DIFF
--- a/pkg/connectors/microcks_client.go
+++ b/pkg/connectors/microcks_client.go
@@ -247,16 +247,25 @@ func (c *microcksClient) GetKeycloakURL() (string, error) {
 		panic(err)
 	}
 
-	// Retrieve auth server url and realm name.
-	enabled := configResp["enabled"].(bool)
-	authServerURL := configResp["auth-server-url"].(string)
-	realmName := configResp["realm"].(string)
-
-	// Return a proper URL or 'null' if Keycloak is disables.
-	if enabled {
-		return authServerURL + "/realms/" + realmName + "/", nil
+	// Retrieve auth server url and realm name using comma-ok assertions
+	// so a malformed response surfaces as a Go error instead of a panic.
+	enabled, ok := configResp["enabled"].(bool)
+	if !ok {
+		return "", fmt.Errorf("Microcks /api/keycloak/config response missing or invalid 'enabled' field")
 	}
-	return "null", nil
+	// Return 'null' if Keycloak is disabled.
+	if !enabled {
+		return "null", nil
+	}
+	authServerURL, ok := configResp["auth-server-url"].(string)
+	if !ok {
+		return "", fmt.Errorf("Microcks /api/keycloak/config response missing or invalid 'auth-server-url' field")
+	}
+	realmName, ok := configResp["realm"].(string)
+	if !ok {
+		return "", fmt.Errorf("Microcks /api/keycloak/config response missing or invalid 'realm' field")
+	}
+	return authServerURL + "/realms/" + realmName + "/", nil
 }
 
 func (c *microcksClient) refreshAuthToken(localCfg *config.LocalConfig, ctxName, configPath string) error {
--- a/pkg/connectors/microcks_client_test.go
+++ b/pkg/connectors/microcks_client_test.go
@@ -102,3 +102,63 @@ func TestDownloadArtifactReturnsResponseBody(t *testing.T) {
 		t.Fatalf("expected response body %q, got %q", expectedBody, msg)
 	}
 }
+
+func TestGetKeycloakURL_Enabled(t *testing.T) {
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.URL.Path != "/api/keycloak/config" {
+			t.Fatalf("unexpected path: %s", r.URL.Path)
+		}
+		if r.Method != http.MethodGet {
+			t.Fatalf("unexpected method: %s", r.Method)
+		}
+		w.Header().Set("Content-Type", "application/json")
+		_, _ = w.Write([]byte(`{"enabled": true, "auth-server-url": "https://kc", "realm": "microcks"}`))
+	}))
+	defer server.Close()
+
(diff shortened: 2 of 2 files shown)


Harishrs2006:
duplicate  #267 rubrx filed the same panic issue earlier ,  root issue is already tracked

hazelmayank:
Sorry I forgot to check if the same issue was addressed before

Issue #403, opened 2026-05-17
CLI crashes with panic if Microcks API returns unexpected responses


PranjaliBhardwaj:
While using the CLI, I noticed that pkg/connectors/microcks_client.go uses panic(err) in several places when reading HTTP response bodies or unmarshaling JSON.

If there is a temporary network glitch (e.g., the connection drops while reading the body) or if the Microcks server returns an invalid JSON response (like a 502 Bad Gateway HTML page from a proxy), the CLI will abruptly crash and print a Go stack trace instead of handling the error gracefully.

Additionally, methods like CreateTestResult use unsafe type assertions on the parsed JSON map (e.g., createTestResp["id"].(string)), which will also cause a panic if the id field is missing from the response payload.

Expected Behavior
The CLI should catch these errors, bubble them up through the standard error return values, and print a clean, user-friendly error message to the terminal before exiting with a non-zero status code.

Actual Behavior
The CLI completely crashes with a stack trace.

Suggested Fix
We should replace all instances of panic(err) in these client methods with proper context-wrapped errors (e.g., fmt.Errorf("failed to read response: %w", err)) since the functions already have an error return type defined. We should also add the comma-ok idiom to the type assertions to ensure we don't crash on missing fields.

Caesarsage:
Thank you

Resolved on master via #489 — the panic(err) sites in pkg/connectors were replaced with classified error returns, so a network blip or a 502 HTML page now surfaces as an actionable error message rather than a stack trace. I believe this can be closed (same underlying report as #478/#401).

PR #404, opened 2026-05-17
Replace unhandled panics with proper error returns in microcks_client

PranjaliBhardwaj:
in pkg/connectors/microcks_client.go, I noticed that several methods were using panic(err) when reading HTTP response bodies or unmarshaling JSON. If the Microcks server returns a bad response or there's a network blip, this would completely crash the CLI.

I went through and swapped out all those panic calls for proper fmt.Errorf returns in the functions that already have an error return type. this way the errors bubble up properly and we can handle them gracefully instead of just dying on the spot.

Implementation details
Removed panic(err) in NewClient, GetKeycloakURL, CreateTestResult, GetTestResult, and UploadArtifact.
Replaced them with context-wrapped errors (e.g. fmt.Errorf("failed to read test result response body: %w", err)).
In CreateTestResult, I also added a safe type assertion check (testID, ok := createTestResp["id"].(string)) so it doesn't panic if the id field happens to be missing or isn't a string.

Testing
Built the CLI locally.
Intentionally caused some network failures/invalid JSON responses and verified that the CLI now prints a clean error message and exits normally instead of throwing a stack trace panic.

### DIFF
--- a/pkg/connectors/microcks_client.go
+++ b/pkg/connectors/microcks_client.go
@@ -151,7 +151,7 @@ func NewClient(opts ClientOptions) (MicrocksClient, error) {
 
 		u, err := url.Parse(apiURL)
 		if err != nil {
-			panic(err)
+			return nil, fmt.Errorf("failed to parse API URL: %w", err)
 		}
 		c.APIURL = u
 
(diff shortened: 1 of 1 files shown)


Caesarsage:
As per https://github.com/microcks/microcks-cli/issues/403#issuecomment-5309150763

This can be closed

Caesarsage:
Thanks for this! The panic-to-error migration landed on master via #489 (which also added error classification), so this PR is superseded. Appreciate the effort — several other issues in the queue are still open if you'd like another target.

PR #405, opened 2026-05-17
fix(login): remove unconditional logging of OAuth2 access and refresh tokens

### LINKED ISSUE #345: fix(login): OAuth2 access and refresh tokens unconditionally leaked to stderr on SSO login
### Describe the bug

during SSO login, `cmd/login.go` unconditionally prints the OAuth2 access token and refresh token to stderr via `log.Printf` after successful authentication — regardless of whether `--verbose` is set.

Lines 296–297 in `cmd/login.go`:
```go

### Expected behavior

token values should never appear in logs or stderr output. The "Authentication successful" message is sufficient confirmation. If token visibility is needed for debugging, it should be gated behind the --verbose flag, consistent with how all other sensitive data is handled in this codebase (via config.Verbose / config.DumpRequestIfRequired).

### Actual behavior

On every SSO login, the full OAuth2 access token string and refresh token string are printed to stderr unconditionally. Any CI/CD pipeline or log aggregation system (GitHub Actions, Jenkins, Datadog, Splunk) that captures stderr will silently record long-lived credentials in plaintext. Refresh tokens are particularly dangerous as they allow obtaining new access tokens without user interaction.

### How to Reproduce?

1) Configure a Microcks instance with Keycloak SSO enabled
2) Run: microcks login --keycloak-realm <realm> --sso (any flow that calls connectAndGetTokenWithSSO in cmd/login.go)
3) Observe that the full token strings are printed to stderr after "Authentication successful"
4) Capture stderr (e.g., redirect to a log file) — credentials are stored in plaintext

All versions including latest (master: b892bbb). Bug introduced with the SSO login flow — affects any release containing cmd/login.go's connectAndGetTokenWithSSO function.

microcks-cli binary (any platform)

### Additional information

The fix is straightforward — remove lines 296–297 from cmd/login.go entirely. The fmt.Printf("Authentication successful\n") line immediately above already confirms success without leaking any credential.

all other sensitive output in the codebase follows the pattern:
if config.Verbose { ... }
or uses config.DumpRequestIfRequired — these two log.Printf calls are the only exception.

A fix PR is available: https://github.com/microcks/microcks-cli/compare/master...Harishrs2006:microcks-cli:fix/oauth-token-leak-login-log

### PR DESCRIPTION
**Pull request by Harishrs2006**
Summary
Removes two `log.Printf` calls in `cmd/login.go` that unconditionally printed the OAuth2 access token and refresh token to stderr on every SSO login — regardless of whether `--verbose` was set.

Problem
```go
// Before — leaks credentials to stderr unconditionally
log.Printf("Token: %s\n", tokenString)
log.Printf("Refresh Token: %s\n", refreshToken)
```
Any CI/CD pipeline or log aggregation system capturing stderr silently records long-lived credentials in plaintext. Refresh tokens are especially dangerous — they allow obtaining new access tokens without user interaction.

All other sensitive output in this codebase is gated behind config.Verbose / config.DumpRequestIfRequired. These two lines were the only exception — leftover debug statements.

Changes:
cmd/login.go: removed 2 lines — the unconditional token log statements
The fmt.Printf("Authentication successful\n") immediately above already confirms success without exposing any credential
Testing
go build ./...  # clean
go test ./...   # all pass
go vet ./...    # clean
Manual: SSO login flow no longer prints token values to stderr. Behavior otherwise unchanged.

Fixes : #345

### DIFF
--- a/cmd/login.go
+++ b/cmd/login.go
@@ -293,8 +293,6 @@ func oauth2login(
 	ctx, cancel := context.WithTimeout(ctx, 1*time.Second)
 	defer cancel()
 	_ = srv.Shutdown(ctx)
-	log.Printf("Token: %s\n", tokenString)
-	log.Printf("Refresh Token: %s\n", refreshToken)
 	return tokenString, refreshToken
 }
 


Harsh4902 (approved):
LGTM

PR #406, opened 2026-05-18
fix(context): guard nil local config before context switch

Rahulatram321:
### Description

This PR fixes a crash path in the `microcks context` command when no local config exists yet.

#### Problem
When running `microcks context <context-name>` on a fresh machine (or after config cleanup), `localCfg` can be `nil`.  
The command dereferences `localCfg` (`localCfg.CurrentContext`) without a nil guard, which can lead to a nil pointer panic instead of a user-friendly message.

#### Changes
- Added a defensive `localCfg == nil` check in `cmd/context.go` before first dereference in the context-switch flow.
- Preserved current behavior for:
  - `microcks context` (no args): still lists contexts / prints existing output path.
  - `microcks context --delete ...`: unchanged.

#### Result
- No panic when config is missing.
- Users now get a clear, deterministic error path instead of a runtime crash.

### DIFF
--- a/cmd/context.go
+++ b/cmd/context.go
@@ -45,6 +45,9 @@ microcks context http://localhost:8080 --delete`,
 				printMicrocksContexts(configPath)
 				return
 			}
+			if localCfg == nil {
+				log.Fatalf("No contexts defined in %s", configPath)
+			}
 
 			ctxName := args[0]
 			if localCfg.CurrentContext == ctxName {
@@ -122,3 +125,4 @@ func printMicrocksContexts(configPath string) {
 		errors.CheckError(err)
 	}
 }
+


Vaishnav88sk:
please pass the DCO

Rahulatram321:
> please pass the DCO

Thanks for the review and for flagging this.

I’ve updated this PR by rewriting the commit(s) with DCO sign-off and force-pushed the branch.
Each commit now includes:

Signed-off-by: Rahul Atram <rahulatram0226@gmail.com>

I also kept the scope unchanged to only the intended fix in this PR.
Could you please recheck when the DCO status refreshes? Thanks again.

Rahulatram321:
Hi Sir,

Thank you for pointing this out.

I have updated this PR by rewriting the commit history with DCO sign-off
and force-pushing the branch.
All commits now include:

Signed-off-by: Rahul Atram ***@***.***>

The PR scope is unchanged and remains limited to the original fix only.
Could you please recheck once the DCO status updates?

Thanks again for your guidance.

Best,
Rahul Atram

<https://www.avast.com/sig-email?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail>
Virus-free.www.avast.com
<https://www.avast.com/sig-email?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail>
<#DAB4FAD8-2DD7-40BB-A1B8-4E2AA1F9FDF2>

Caesarsage:
Thanks! The context command on master now nil-guards the local config before use, so this is superseded. Appreciate the effort.

Suggest closing this

Harsh4902:
Thanks @Rahulatram321 and sorry we couldn't merge this erlier. As it's already tackled on master, we have to close this.

PR #420, opened 2026-05-23
fix: logout resolves named contexts

### LINKED ISSUE #419: bug: logout cannot handle named contexts despite documented usage
## Summary

`microcks logout CONTEXT` is documented as supporting a named context, but the implementation passes the provided argument directly to `LocalConfig.RemoveToken`. That function searches `LocalConfig.Users` by user name, while `login --name <name>` stores the user under the server URL and stores the friendly name only in `ContextRef.Name`.

As a result, logging in with a named context works, but logging out by that same context name fails and leaves the auth and refresh tokens in the local config.

## Evidence from the code

- `cmd/login.go` stores the user with `Name: server`, then stores the named context with `Name: ctxName` and `User: server`.
- `cmd/logout.go` reads `context := args[0]` and calls `localCfg.RemoveToken(context)` directly.
- `pkg/config/localconfig.go` implements `RemoveToken` by comparing `u.Name == serverName`, so a context alias such as `staging` will not match the stored user name such as `https://microcks.example`.

## Reproduction

1. Log in with a named context:

```bash
microcks login https://microcks.example --name staging
```

2. Try to log out using the documented named context form:

```bash
microcks logout staging
```

## Actual behavior

The command fails with a message like:

```text
Context staging does not exist
```

The token entry stored under the server URL remains in the config file.

## Expected behavior

`microcks logout staging` should resolve the context name first, find the referenced user/server entry, and clear the associated auth and refresh tokens. If the CLI also wants to keep accepting raw server URLs, it can support both forms, but the documented named-context form should work.

## Impact

This affects a core authentication workflow. Users who use named contexts can believe they are logging out, but the CLI cannot clear their tokens unless they know to pass the underlying server URL instead of the context name.

gyanranjanpanda:
## Summary

Fixes #419.

`microcks logout` now resolves the provided argument as a Microcks context before clearing tokens. That makes the documented named-context flow work for logins created with `microcks login ... --name <context>`, while still preserving the old behavior of accepting the stored user/server name directly.

## Root cause

`login --name staging` stores the user under the server URL and stores `staging` only as the context name. `logout` passed the argument directly to `RemoveToken`, which searches users by name, so `microcks logout staging` could not find the token entry.

## Validation

- `go test ./cmd -run TestLogoutContext -count=1`
- `go test ./pkg/config ./pkg/connectors`
- `go vet ./cmd`

Note: `go test ./...` is still blocked by the existing `TestDeleteContext` fixture issue where `cmd/context_test.go` writes `./testdata/local.config` before creating `cmd/testdata`.

### DIFF
--- a/cmd/logout.go
+++ b/cmd/logout.go
@@ -7,7 +7,6 @@ import (
 
 	"github.com/microcks/microcks-cli/pkg/config"
 	"github.com/microcks/microcks-cli/pkg/connectors"
-	"github.com/microcks/microcks-cli/pkg/errors"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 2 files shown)


Caesarsage:
Thank you @gyanranjanpanda 


Fixed on master — logout now resolves the argument via ResolveContext before clearing tokens, so the documented named-context flow works (covered by cmd/logout_test.go). 

I believe this can be closed. 

Thanks for the precise root-cause analysis here — it matches what was implemented.

gyanranjanpanda:
thanks @Caesarsage i will close this

PR #422, opened 2026-05-23
fix: keep shared server config when deleting one of multiple contexts

### LINKED ISSUE #421: Deleting one context removes config when multiple contexts share a single server
### Describe the bug

  If the local config contains multiple contexts for the same Microcks server,
  deleting one of those contexts can remove the shared server config for all of them.

  ### To Reproduce

  1. Start Microcks locally:
     ```bash
     ./microcks start

  2. Create another context pointing to the same server:

     ./microcks login http://localhost:8585 --name dev
  3. Delete one of the contexts:

     ./microcks context dev --delete
  4. Inspect the local config:

     cat ~/.config/microcks/config

  ### Current behavior

  When there is only one server entry in the config and multiple contexts point to
  that same server, deleting one context can also remove the shared server-related
  metadata. In some cases, the whole config file is deleted(including the metada of the isntance).

  


### Expected behavior

### Expected behavior

  Deleting one context should remove only that context. The shared server, user, auth,
  and instance metadata should remain as long as another context still references the
  same server.

  ### Additional context

  This was reproduced with multiple contexts pointing to http://localhost:8585.



### Actual behavior

_No response_

### How to Reproduce?

_No response_

_No response_

_No response_

### Additional information

_No response_

Syedowais312:
### Summary

  This fixes context deletion when multiple contexts reference the same server entry.

  If the config contains a single server and multiple contexts pointing to it,
  deleting one context could also remove shared metadata that was still needed by the
  remaining context.

  ### Changes

  - resolve the target context before deleting it
  - delete only the selected context
  - remove shared `user`, `server`, `auth`, and `instance` entries only when no
  remaining context references them
  - add a regression test for multiple contexts sharing one server

  ### Reproducer

  ```bash
  ./microcks start
  ./microcks login http://localhost:8585 --name dev
  ./microcks context dev --delete
```
```bash
❯ cat ~/.config/microcks/config
cat: /home/syedowais/.config/microcks/config: No such file or directory

```

  ### Testing
```bash
  go test ./cmd ./pkg/config
```

closes: https://github.com/microcks/microcks-cli/issues/421

### DIFF
--- a/cmd/context.go
+++ b/cmd/context.go
@@ -72,12 +72,25 @@ func deleteContext(context, configPath string) error {
 	if localCfg == nil {
 		return fmt.Errorf("Nothing to logout from")
 	}
-	serverName, ok := localCfg.RemoveContext(context)
+	resolvedContext, err := localCfg.ResolveContext(context)
+	if err != nil {
+		return fmt.Errorf("Context %s does not exist", context)
+	}
+	_, ok := localCfg.RemoveContext(context)
 	if !ok {
 		return fmt.Errorf("Context %s does not exist", context)
 	}
-	_ = localCfg.RemoveUser(context)
-	_ = localCfg.RemoveServer(serverName)
+
+	if !localCfg.HasContextUsingUser(resolvedContext.User.Name) {
+		_ = localCfg.RemoveUser(resolvedContext.User.Name)
+	}
+	if !localCfg.HasContextUsingServer(resolvedContext.Server.Server) {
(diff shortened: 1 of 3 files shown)


Caesarsage:
Thank you @Syedowais312 

Fixed on master — deleteContext now reference-counts: the server (and user) entries are removed only when no remaining context references them, so shared server config survives deleting one of several contexts. 

I believe this can be closed.

PR #435, opened 2026-05-25
fix: continue importing remaining URLs on failure in import-url command

### LINKED ISSUE #399: fix: import-url command exits early on first failure when importing multiple URLs
Why we need this:
When importing multiple comma-separated URLs, if one URL fails,
the command immediately exits and remaining URLs are never imported.
This is the same issue that was fixed in the import command in PR #394.

How this will help:
All URLs should be attempted regardless of individual failures,
giving users complete feedback in one run.

Motivation:
Consistent behavior across all import commands (import, import-url).

## Problem
When running:
microcks import-url "https://url1.yaml,https://url2.yaml,https://url3.yaml"

If url1 fails, os.Exit(1) is called immediately inside the loop
and url2, url3 are never imported.

Looking at cmd/importURL.go:
for _, f := range sepSpecificationFiles {
    msg, err := mc.DownloadArtifact(f, mainArtifact, secret)
    if err != nil {
        os.Exit(1)  // exits immediately, remaining URLs skipped
    }
}

Same bug as #393, fixed in import command via PR #394.

## Fix
- Collect errors instead of exiting immediately
- Attempt all URLs regardless of individual failures
- Exit with code 1 only if any URL failed

A-d-i-t-y:
## Summary
Fix early exit bug in import-url command when importing multiple URLs.

## Problem
When running:
microcks import-url "https://url1.yaml,https://url2.yaml,https://url3.yaml"

If url1 fails, os.Exit(1) is called immediately and url2, url3 
are never imported. Same bug as #393 which was fixed in import 
command via PR #394.

## Fix
- Replaced os.Exit(1) with error collection and continue
- All URLs are attempted regardless of individual failures
- Exit with code 1 only if any URL failed

## Example output after fix
Failed to import 'https://url2.yaml': connection refused
Microcks has discovered 'https://url1.yaml'
Microcks has discovered 'https://url3.yaml'

## Files Changed
- cmd/importURL.go — replaced os.Exit(1) with hasErrors flag and continue

Closes #399

### DIFF
--- a/cmd/importURL.go
+++ b/cmd/importURL.go
@@ -95,6 +95,7 @@ func NewImportURLCommand(globalClientOpts *connectors.ClientOptions) *cobra.Comm
 					return
 				}
 			}
+			var hasErrors bool
 			sepSpecificationFiles := strings.Split(specificationFiles, ",")
 			for _, f := range sepSpecificationFiles {
 				mainArtifact := true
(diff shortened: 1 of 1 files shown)


A-d-i-t-y:
Hi @Harsh4902 @Vaishnav88sk, I have raised this PR to fix the early exit bug in import-url command as discussed in issue #399. Same fix as PR #394 for the import command - all URLs are now attempted regardless of individual failures. Would love your feedback!

PR #436, opened 2026-05-25
fix: clean up auth when deleting named contexts

### LINKED ISSUE #427: bug: context --delete leaves stale auth tokens for named contexts
### Describe the bug

When deleting a context with `microcks context <name> --delete`, the CLI removes the context and server entries but does not remove the associated user credentials when the context was created with a custom name via `login --name`.

`login` stores users under the server URL (`users[].name`), while contexts use the context name (`contexts[].name`) and reference the user through `contexts[].user`.

However, `deleteContext` calls `RemoveUser(context)` using the context name. For a named context like `dev`, the actual user entry is keyed by `http://localhost:8080`, so cleanup silently fails.

This can leave OAuth tokens and related auth records in `~/.config/microcks/config` even after the user is told `Context 'dev' deleted`.

### Expected behavior

After `microcks context dev --delete`:

- The `dev` context entry is removed from `contexts`
- The related `servers`, `users` auth-token / refresh-token, and `auths` entries for that Microcks instance are removed
- No OAuth tokens or Keycloak client credentials remain on disk for that deleted context

### Actual behavior

After deleting a named context:

- The `contexts` entry is removed
- The related `servers` entry is removed
- The related `users` entry remains, including `auth-token` and `refresh-token`
- The CLI prints `Context 'dev' deleted` with no warning that credentials were left behind

### How to Reproduce?

Reproducer:

1. Create a temporary config:

```bash
mkdir -p /tmp/microcks-cli-verify

cat > /tmp/microcks-cli-verify/config <<'EOF'
current-context: staging
contexts:
  - name: dev
    server: http://localhost:8080
    user: http://localhost:8080
    instance: ""
  - name: staging
    server: http://localhost:8083
    user: http://localhost:8083
    instance: ""
servers:
  - server: http://localhost:8080
    insecureTLS: true
    keycloakEnable: true
  - server: http://localhost:8083
    insecureTLS: true
    keycloakEnable: true
users:
  - name: http://localhost:8080
    auth-token: LEAKED-TOKEN
    refresh-token: LEAKED-REFRESH
  - name: http://localhost:8083
    auth-token: ""
    refresh-token: ""
auths:
  - server: http://localhost:8080
    clientId: my-client
    clientSecret: my-secret
EOF

chmod 600 /tmp/microcks-cli-verify/config
```

2. Delete the named context:
```bash
./build/dist/microcks context dev --delete --config /tmp/microcks-cli-verify/config
```

3. Inspect the config:
```bash
cat /tmp/microcks-cli-verify/config
```

Image

Microcks-CLI 1.0.2

_No response_

### Additional information

_No response_

cotishq:
### Description

- Fixes named context deletion so cleanup uses the removed context's referenced `user` and `server` values instead of the context name.
- Removes the associated auth entry when deleting a context, matching the expected cleanup behavior for credentials tied to that server.
- Adds a regression test covering `context --delete` for a named context where `contexts[].name` differs from `contexts[].user`.

### Screenshot
Screenshot From 2026-05-25 12-04-04

### Related issue(s)

Fixes #427

### DIFF
--- a/cmd/context.go
+++ b/cmd/context.go
@@ -73,12 +73,13 @@ func deleteContext(context, configPath string) error {
 	if localCfg == nil {
 		return fmt.Errorf("Nothing to logout from")
 	}
-	serverName, ok := localCfg.RemoveContext(context)
+	contextRef, ok := localCfg.RemoveContext(context)
 	if !ok {
 		return fmt.Errorf("Context %s does not exist", context)
 	}
-	_ = localCfg.RemoveUser(context)
-	_ = localCfg.RemoveServer(serverName)
+	_ = localCfg.RemoveUser(contextRef.User)
+	_ = localCfg.RemoveServer(contextRef.Server)
+	_ = localCfg.RemoveAuth(contextRef.Server)
 
 	if localCfg.IsEmpty() {
 		err := localCfg.DeleteLocalConfig(configPath)
--- a/cmd/context_test.go
+++ b/cmd/context_test.go
@@ -65,8 +65,56 @@ func TestDeleteContext(t *testing.T) {
 	require.NoError(t, err)
 }
 
-func TestDeleteContextEmpty(t *testing.T) {
-	err := deleteContext("http://localhost:8080", "./testdata/non-existent-file.config")
-	require.EqualError(t, err, "Nothing to logout from")
-}
+func TestDeleteNamedContextRemovesReferencedUserAndAuth(t *testing.T) {
+	configPath := t.TempDir() + "/config"
+	localConfig := `current-context: staging
+contexts:
+- name: dev
+  server: http://localhost:8080
+  user: http://localhost:8080
+  instance: ""
+- name: staging
+  server: http://localhost:8083
+  user: http://localhost:8083
+  instance: ""
+servers:
+- name: ""
+  server: http://localhost:8080
+  insecureTLS: true
+  keycloakEnable: true
+- name: ""
+  server: http://localhost:8083
+  insecureTLS: true
+  keycloakEnable: true
+users:
+- name: http://localhost:8080
+  auth-token: stale-token
+  refresh-token: stale-refresh-token
+- name: http://localhost:8083
+  auth-token: ""
+  refresh-token: ""
+auths:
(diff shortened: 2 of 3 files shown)


cotishq:
hey @Harsh4902 , would love your feedback here

Vaishnav88sk:
@cotishq Please resolve the conflicts.

Vaishnav88sk:
The current delete path uses the context name for user cleanup, which is wrong for named contexts where `contexts[].name != contexts[].user`. Cleaning up by the removed `ContextRef` is the right direction, and adding auth cleanup for the associated server also makes sense.

One thing to double-check before merge: are `server`, `user`, or `auth` entries ever intentionally shared by multiple contexts? If yes, deleting them unconditionally here could remove credentials/config still referenced by another context. If they are always 1:1 with a context, then this is fine.

If shared references are not a supported case, LGTM. @Harsh4902

cotishq:
yeah you are right, since microcks login supports the --name flag, users can define multiple named contexts pointing to the same server or user. so if one is deleted, deleting the server, user or auth config unconditionally would indeed break the remaining contexts that reference them.

i think we should check if any other contexts still reference the removed contexts user or server before deleting them.

what's your opinion should i update the PR to add references checking ?

Vaishnav88sk:
> yeah you are right, since microcks login supports the --name flag, users can define multiple named contexts pointing to the same server or user. so if one is deleted, deleting the server, user or auth config unconditionally would indeed break the remaining contexts that reference them.
> 
> i think we should check if any other contexts still reference the removed contexts user or server before deleting them.
> 
> what's your opinion should i update the PR to add references checking ?

Yes, please update the PR with reference checks.

Because multiple named contexts can share the same `server` / `user` / `auth`, we should only remove those entries if the deleted context was the last one referencing them.

Please add tests for both shared-reference and last-reference cases as well. Then I’ll re-review.

Caesarsage:
Thanks @cotishq 

This has been fixed on master , deleteContext now resolves the context first and removes the referenced user (and its tokens) via the context's own user reference, not the context name, and only when no other context still references it. 

I believe this can be closed.

PR #439, opened 2026-05-25
fix: add --name flag to stop command and remove empty println

### LINKED ISSUE #438: fix: stop command missing --name flag to specify which instance to stop
Why we need this:
The start command supports --name flag to name a Microcks instance
but the stop command has no --name flag. Users cannot specify which
instance to stop when multiple instances exist.

How this will help:
Adding --name flag to stop command makes it consistent with start
command and allows users to stop a specific named instance.

## Problem
microcks start --name myinstance  ✅ works
microcks stop --name myinstance   ❌ flag does not exist

The stop command always resolves the current context instead of
allowing the user to specify which instance to stop by name.

## Fix
- Add --name flag to stop command
- Pass name to ResolveContext() so correct instance is stopped
- Remove unnecessary empty fmt.Println() call

A-d-i-t-y:
## Summary
Adds --name flag to stop command for consistency with start command.

## Problem
The start command supports --name flag to name a Microcks instance
but the stop command has no --name flag:

microcks start --name myinstance  ✅ works
microcks stop --name myinstance   ❌ flag did not exist

Users could not specify which instance to stop when multiple 
instances exist. The stop command always resolved the current 
context instead.

## Fix
- Added --name flag to stop command
- Pass name to ResolveContext() so correct instance is stopped
- Removed unnecessary empty fmt.Println() call

## Files Changed
- cmd/stop.go — added --name flag, removed empty println

Closes #438

### DIFF
--- a/cmd/stop.go
+++ b/cmd/stop.go
@@ -12,6 +12,7 @@ import (
 
 func NewStopCommand(globalClientOpts *connectors.ClientOptions) *cobra.Command {
 
+	var name string
 	var stopCmd = &cobra.Command{
 		Use:   "stop",
 		Short: "stop microcks instance",
@@ -27,7 +28,7 @@ func NewStopCommand(globalClientOpts *connectors.ClientOptions) *cobra.Command {
 				return
 			}
 
-			ctx, err := localConfig.ResolveContext("")
+			ctx, err := localConfig.ResolveContext(name)
 			errors.CheckError(err)
 			instance := ctx.Instance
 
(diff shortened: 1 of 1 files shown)


A-d-i-t-y:
Hi @Harsh4902 @Vaishnav88sk, I have raised this PR to add --name flag to the stop command. Currently start command supports --name flag but stop command did not, making it inconsistent. This fix allows users to specify which Microcks instance to stop by name. Would love your feedback!

Vaishnav88sk:
Please check #307

It may cause conflict.

A-d-i-t-y:
Hi @Vaishnav88sk, thank you for checking! I can see PR #307 already covers significant improvements to the stop command. Closing this PR to avoid conflicts. Sorry for the noise!

PR #451, opened 2026-05-29
feat: add --client-id and --client-secret flags to login command

A-d-i-t-y:
## Summary
Adds --client-id and --client-secret flags to the login command 
so client credentials can be passed directly without requiring 
environment variables.

## Problem
Currently the login command only reads client credentials from 
environment variables:

MICROCKS_CLIENT_ID=my-client
MICROCKS_CLIENT_SECRET=my-secret
microcks login http://localhost:8080 --username admin --password secret

This is inconvenient in CI/CD environments where injecting env vars 
is not always possible or desirable.

## Solution
Added two new flags with environment variables as fallback:

microcks login http://localhost:8080 \
  --username admin \
  --password secret \
  --client-id my-client \
  --client-secret my-secret

If flags are not provided, environment variables are still used 
as fallback — no breaking changes.

## Files Changed
- cmd/login.go — added --client-id and --client-secret flags

### DIFF
--- a/cmd/login.go
+++ b/cmd/login.go
@@ -31,6 +31,8 @@ func NewLoginCommand(globalClientOpts *connectors.ClientOptions) *cobra.Command 
 		ctxName          string
 		username         string
 		password         string
+		clientID         string
+		clientSecret     string
 		sso              bool
 		ssoLaunchBrowser bool
 		ssoProt          int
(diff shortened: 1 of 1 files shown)


A-d-i-t-y:
Hi @Harsh4902 @Vaishnav88sk @lbroudoux, I have raised this PR to add --client-id and --client-secret flags to the login command. Previously credentials could only be provided via environment variables which is inconvenient in many CI/CD setups. This change adds flag support with environment variables as fallback — no breaking changes. Would love your feedback!

Vaishnav88sk:
@A-d-i-t-y From onwards, please create an issue first, and illustrate the actual problem with reproduction steps (if suitable).

Vaishnav88sk:
Before merge, I’d like two small follow-ups:
- update the command help/examples to show the new flags explicitly, (Check if necessary)
- add a small test for flag-vs-env behavior (flags provided, env fallback, and flags taking precedence over env vars).

One note to keep in mind in the docs/help: passing secrets via CLI flags is convenient, but it is also more visible in shell history/process lists.

PR #455, opened 2026-06-11
fix(start): wait for Microcks server readiness before returning

Caesarsage:
### Description

microcks start returns as soon as the container process is up: ~20–40s before the Microcks server inside actually serves traffic, so chained microcks import/test calls intermittently fail with connection-refused.

This polls GET /api/keycloak/config until it returns 200 before printing success. 

New flags: --ready-timeout (default 60s) and --no-wait for fire-and-forget. 

Unit tests cover the ready, retry, and timeout paths.

### Related issue(s)

 https://github.com/microcks/microcks-cli/issues/255

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -3,6 +3,8 @@ package cmd
 import (
 	"fmt"
 	"log"
+	"net/http"
+	"time"
 
 	"github.com/microcks/microcks-cli/pkg/config"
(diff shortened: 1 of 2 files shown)


Caesarsage:
cc 
@Harsh4902 @Vaishnav88sk

Harsh4902 (approved):
LGTM

PR #461, opened 2026-06-17
fix(stop): add --name flag to stop named contexts and instances

### LINKED ISSUE #438: fix: stop command missing --name flag to specify which instance to stop

AdeshDeshmukh:
## Problem

The `microcks stop` command only supports stopping the instance associated with the *current context*. There is no way to target a specific instance by name — even though `microcks start --name <name>` and `microcks login --name <name>` already allow users to create named contexts and instances.

This asymmetry means that after starting multiple instances or logging into multiple servers, users have no direct way to stop a specific one without first switching context.

## Root Cause

When a user runs `microcks start --name myinstance`, the CLI stores the context with:
- **Context name** = `http://localhost:8585` (the server URL)
- **Instance field** = `myinstance` (the user-given name)

So a naive lookup by context name alone (`ResolveContext(myinstance)`) fails — there is no context literally named `myinstance`. The existing `ResolveContext` expects the context's *name* field, not the Instance field.

## Solution

This PR adds a `--name` flag to `microcks stop` with a **two-tier fallback resolution**:

1. **Try as a context name** — works for contexts created via `microcks login --name dev` (stored as `Context{Name: "dev"}`)
2. **Scan contexts by Instance field** — works for contexts created via `microcks start --name myinstance` (stored as `ContextRef{Name: "http://...", Instance: "myinstance"}`)

If neither lookup succeeds, the user receives a clear error message.

When `--name` is omitted, behavior is 100% identical to the current code (uses the current context).

## Changes

`cmd/stop.go` — 29 insertions, 2 deletions:
- Added `var name string` for the flag variable
- Added `Example` help text with usage patterns
- Added two-tier context resolution logic
- Registered the `--name` flag on `stopCmd`

## Verification

| Check | Result |
|-------|--------|
| `go build ./...` | ✅ Compiles cleanly |
| All existing tests | ✅ Pass (26/27 pass; TestDeleteContext failure is pre-existing) |
| Backward compatible | ✅ No flag = unchanged behavior |

## Example Usage

```bash
# Stop the instance from the current context (existing behavior)
microcks stop

# Stop by context name (useful after login --name)
microcks stop --name dev-context

# Stop by instance name (useful after start --name)
microcks stop --name myinstance
```

Fixes #438

### DIFF
--- a/cmd/stop.go
+++ b/cmd/stop.go
@@ -11,11 +11,17 @@ import (
 )
 
 func NewStopCommand(globalClientOpts *connectors.ClientOptions) *cobra.Command {
+	var name string
 
 	var stopCmd = &cobra.Command{
 		Use:   "stop",
 		Short: "stop microcks instance",
 		Long:  "stop microcks instance",
+		Example: `# Stop the instance from the current context
+microcks stop
+
+# Stop by context name or instance name
+microcks stop --name myinstance`,
 		Run: func(cmd *cobra.Command, args []string) {
 
 			configFile := globalClientOpts.ConfigPath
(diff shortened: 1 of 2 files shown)


Vaishnav88sk:
Ptal at Copilot comments.

If `start --name` and named login contexts exist, `stop` should also let users target them directly.

The resolution approach looks reasonable, but before merge I’d like targeted tests for:
- resolving by context name,
- resolving by instance name,
- not-found behavior.

I’d also like the precedence to be explicit and tested when a context name and an instance name collide. Right now context lookup wins, which is fine, but it should be intentional.

PR #464, opened 2026-06-18
fix (cmd,pkg) : Potential Panic/Crash via Nil Pointer Dereference in Switch Context Command

### LINKED ISSUE #462: Potential Panic/Crash via Nil Pointer Dereference in Switch Context Command
### Where
* **File:** [cmd/context.go](file:///c:/Users/Hp/microcks-cli/cmd/context.go#L32-L57)
* **Function:** `NewContextCommand`

### Description
If a user runs the `context` command on a fresh installation without a pre-existing configuration, `config.ReadLocalConfig(...)` returns `nil, nil`. The command checks for errors but not for `nil` values, and proceeds directly to access `localCfg.CurrentContext`:
```go
localCfg, err := config.ReadLocalConfig(configPath)
errors.CheckError(err)
...
if localCfg.CurrentContext == ctxName {
```
This causes a runtime nil-pointer dereference panic.

### Reason ?
Standard CLI usability dictates that commands should fail gracefully with a user-friendly message rather than panicking when no configuration is initialized.

### Bug Simulation Workflow
```mermaid
graph TD
    A[Fresh installation: config file does not exist] --> B[User runs: microcks context dev-server]
    B --> C[ReadLocalConfig returns nil, nil]
    C --> D[CheckError passes since err is nil]
    D --> E[Evaluate localCfg.CurrentContext]
    E --> F{Panic: runtime error: invalid memory address or nil pointer dereference}
```

#### Before (Lines to be changed)
```diff
-			localCfg, err := config.ReadLocalConfig(configPath)
-			errors.CheckError(err)
-			if delete {
```

#### After (Changed lines)
```diff
+			localCfg, err := config.ReadLocalConfig(configPath)
+			errors.CheckError(err)
+			if localCfg == nil {
+				log.Fatalf("No contexts defined. Please login first.")
+			}
+			if delete {
```

---

aniket866:
### Description
If a user runs the `context` command on a fresh installation without a pre-existing configuration, `config.ReadLocalConfig(...)` returns `nil, nil`. The command checks for errors but not for `nil` values, and proceeds directly to access `localCfg.CurrentContext`:
```go
localCfg, err := config.ReadLocalConfig(configPath)
errors.CheckError(err)
...
if localCfg.CurrentContext == ctxName {
```
This causes a runtime nil-pointer dereference panic.

### Reason 
Standard CLI usability dictates that commands should fail gracefully with a user-friendly message rather than panicking when no configuration is initialized.

image

Closes #462

### DIFF
--- a/cmd/context.go
+++ b/cmd/context.go
@@ -47,6 +47,7 @@ microcks context http://localhost:8080 --delete`,
 			}
 
 			ctxName := args[0]
+			errors.CheckConfigNil(localCfg == nil, configPath)
 			if localCfg.CurrentContext == ctxName {
 				fmt.Printf("Already at context '%s'\n", localCfg.CurrentContext)
 				return
@@ -100,9 +101,7 @@ func deleteContext(context, configPath string) error {
 func printMicrocksContexts(configPath string) {
 	localCfg, err := config.ReadLocalConfig(configPath)
 	errors.CheckError(err)
-	if localCfg == nil {
-		log.Fatalf("No contexts defined in %s", configPath)
-	}
+	errors.CheckConfigNil(localCfg == nil, configPath)
 	w := tabwriter.NewWriter(os.Stdout, 0, 0, 2, ' ', 0)
 	defer func() { _ = w.Flush() }()
 	columnNames := []string{"CURRENT", "NAME", "SERVER"}
--- a/cmd/context_test.go
+++ b/cmd/context_test.go
@@ -64,3 +64,9 @@ func TestDeleteContext(t *testing.T) {
 	_, err = config.ReadLocalConfig(testConfigFilePath)
 	require.NoError(t, err)
 }
+
+func TestDeleteContextEmpty(t *testing.T) {
+	err := deleteContext("http://localhost:8080", "./testdata/non-existent-file.config")
+	require.EqualError(t, err, "Nothing to logout from")
+}
+
--- a/pkg/errors/error.go
+++ b/pkg/errors/error.go
@@ -24,6 +24,13 @@ func CheckError(err error) {
 	}
 }
 
+func CheckConfigNil(isNil bool, path string) {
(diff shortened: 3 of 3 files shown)


aniket866:
@lbroudoux @Harsh4902 
Please review whenver you get time

pkg/connectors/container_client.go
```diff
@@ -13,7 +13,6 @@ import (
 	"github.com/docker/docker/client"
 	"github.com/docker/docker/pkg/jsonmessage"
 	"github.com/docker/go-connections/nat"
-	"github.com/microcks/microcks-cli/pkg/errors"
```
Harsh4902, line 16: As this PR is to panic error in `context` command. Is there any specific reason to update container client?

pkg/connectors/container_client.go
```diff
@@ -69,21 +68,30 @@ func NewPodmanClient() (*containerClient, error) {
 	switch osName {
 	case "windows":
 		remoteSocket, err := exec.Command("podman", "machine", "inspect", "--format", "{{.ConnectionInfo.PodmanPipe.Path}}").Output()
-		errors.CheckError(err)
```
Harsh4902, line 72: As we have built, a shared error package to can just do improvements in that package rather than removing it and introducing error checks at each file?

pkg/connectors/container_client.go
aniket866: hi @Harsh4902 I have fixed that accidental changes

pkg/connectors/container_client.go
aniket866: fixed as required

aniket866:
hi @Harsh4902 I have done the required changes , now PR is ready to be reviewed, do whenever you get time

Harsh4902 (approved):
LGTM. Thanks @aniket866

PR #456, opened 2026-06-11
fix(start): reconcile instance status against container runtime before trusting it

Caesarsage:
### Description
microcks start trusts the cached status: Running in ~/.config/microcks/config. If the container was removed externally (system restart, autoRemove, manual docker rm), the CLI wedges in "already running" with no recovery path.

This adds ContainerExists() to the container client and checks it before trusting the recorded status — a missing container now falls through to the fresh-create path. Also removes a duplicated "Running" check.

Reproduce: 

- `microcks start --name demo`
- `docker rm -f demo` 
- `microcks start --name demo`
- previously "already running", 
- now recreates the container.

### Related issue(s)
An issue I ran into while experimenting with VS Code extension on top of the CLI.
LFX Mentorship
https://github.com/microcks/microcks-cli/issues/255

PR #468, opened 2026-06-22
fix: remove stale instance entry on recreate and guard stop against missing containers

### LINKED ISSUE #483: bug: stale instance entries accumulate in config when container is recreated
## Description

When a container is removed externally (e.g. `docker stop && docker rm`) and
`microcks start` recreates it, the old instance entry is never removed from config.
A new entry is appended instead, causing duplicate instance records to accumulate.

### Expected behavior

_No response_

### Actual behavior

_No response_

## Steps to Reproduce

1. `./microcks start`
2. `docker stop microcks && docker rm microcks`
3. `./microcks start`
4. `cat ~/.config/microcks/config`  two instance entries with same name but different container IDs

## Screenshots

Image

Image

## Root Cause

`RemoveInstance` matched by `Name` instead of `ContainerID`. Since the recreated
container has a new ID, `UpsertInstance` never finds a match and always appends.

_No response_

_No response_

### Additional information

_No response_

Syedowais312:
## Problem

When a container is removed externally (e.g. `docker system prune`) and
`microcks start` recreates it, the old instance entry was never removed from
config  causing duplicate instance records to accumulate on every recreate.

Two root causes:
1. `RemoveInstance` matched by `Name` (not unique) instead of `ContainerID` (unique)
2. The `autoRemove` path in `stop.go` was passing `instance.Name` to `RemoveInstance`,
   silently failing to remove the correct entry
3. `start.go` was clearing `instance.ContainerID = ""` instead of calling
   `RemoveInstance` to actually remove the stale record from config

## Changes

- `cmd/start.go`: call `localConfig.RemoveInstance(instance.ContainerID)` to
  remove the stale record before recreating, instead of just clearing the local variable
- `cmd/stop.go`: pass `instance.ContainerID` to `RemoveInstance` in the `autoRemove` path
- `pkg/config/localconfig.go`: `RemoveInstance` now matches by `ContainerID` instead of `Name`

## Testing

Ran `docker rm <container>` while config shows `status: Running`, then
`microcks start --name <instance>` config shows a single clean entry with
the new container ID, no duplicates.

Realted: https://github.com/microcks/microcks-cli/pull/456

Fix: https://github.com/microcks/microcks-cli/issues/483

### DIFF
--- a/cmd/stop.go
+++ b/cmd/stop.go
@@ -59,7 +59,7 @@ func NewStopCommand(globalClientOpts *connectors.ClientOptions) *cobra.Command {
 				_ = localConfig.RemoveServer(ctx.Server.Server)
 				_ = localConfig.RemoveUser(ctx.User.Name)
 				_ = localConfig.RemoveAuth(ctx.Server.Server)
-				_ = localConfig.RemoveInstance(instance.Name)
+				_ = localConfig.RemoveInstance(instance.ContainerID)
 
 				localConfig.CurrentContext = ""
 				log.Printf("Instance %s removed successfully", instance.Name)
--- a/cmd/stop.go
+++ b/cmd/stop.go
@@ -40,6 +40,14 @@ func NewStopCommand(globalClientOpts *connectors.ClientOptions) *cobra.Command {
 			errors.CheckError(err)
 			defer containerClient.CloseClient()
 
+			exists, _ := containerClient.ContainerExists(instance.ContainerID)
+			errors.CheckError(err)
+			if !exists {
+				fmt.Printf("Container for instance %s no longer exists\n", instance.Name)
+				fmt.Printf("Run 'microcks start --name %s' to bring the container back\n", instance.Name)
+				return
+			}
+
 			err = containerClient.StopContainer(instance.ContainerID)
 			if err != nil {
 				log.Fatalf("Failed to stop a container: %v", err)
@@ -59,7 +67,7 @@ func NewStopCommand(globalClientOpts *connectors.ClientOptions) *cobra.Command {
 				_ = localConfig.RemoveServer(ctx.Server.Server)
 				_ = localConfig.RemoveUser(ctx.User.Name)
 				_ = localConfig.RemoveAuth(ctx.Server.Server)
-				_ = localConfig.RemoveInstance(instance.ContainerID)
+				_ = localConfig.RemoveInstance(instance.Name)
 
 				localConfig.CurrentContext = ""
 				log.Printf("Instance %s removed successfully", instance.Name)
--- a/cmd/stop.go
+++ b/cmd/stop.go
@@ -40,7 +40,7 @@ func NewStopCommand(globalClientOpts *connectors.ClientOptions) *cobra.Command {
 			errors.CheckError(err)
 			defer containerClient.CloseClient()
 
-			exists, _ := containerClient.ContainerExists(instance.ContainerID)
+			exists, err := containerClient.ContainerExists(instance.ContainerID)
 			errors.CheckError(err)
 			if !exists {
 				fmt.Printf("Container for instance %s no longer exists\n", instance.Name)
@@ -67,7 +67,7 @@ func NewStopCommand(globalClientOpts *connectors.ClientOptions) *cobra.Command {
 				_ = localConfig.RemoveServer(ctx.Server.Server)
 				_ = localConfig.RemoveUser(ctx.User.Name)
 				_ = localConfig.RemoveAuth(ctx.Server.Server)
-				_ = localConfig.RemoveInstance(instance.Name)
+				_ = localConfig.RemoveInstance(instance.ContainerID)
 
 				localConfig.CurrentContext = ""
 				log.Printf("Instance %s removed successfully", instance.Name)
--- a/cmd/stop.go
+++ b/cmd/stop.go
@@ -62,6 +62,16 @@ func NewStopCommand(globalClientOpts *connectors.ClientOptions) *cobra.Command {
 			}
 			defer containerClient.CloseClient()
 
+			exists, err := containerClient.ContainerExists(instance.ContainerID)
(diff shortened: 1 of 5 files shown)


Vaishnav88sk:
Matching by `ContainerID` instead of `Name` is the correct approach. Could you add a test case for the duplicate-entry scenario?

Syedowais312:
> Matching by `ContainerID` instead of `Name` is the correct approach. Could you add a test case for the duplicate-entry scenario?

i have added  appropriate test cases in pkg/config/localconfig_test.go

pkg/config/localconfig_test.go
Vaishnav88sk, line 1: Please add the standard Apache license header to this new test file for consistency with the rest of the repository.

Vaishnav88sk:
Thanks for adding tests after the earlier review feedback.

One thing I want to call out: the PR title mentions “guard stop against missing containers”, but I don’t see an actual missing-container guard added in `stop.go` in this diff. So:
- add the missing guard behavior as well.

Syedowais312:
> Thanks for adding tests after the earlier review feedback.
> 
> One thing I want to call out: the PR title mentions “guard stop against missing containers”, but I don’t see an actual missing-container guard added in `stop.go` in this diff. So:


sorry i missed it earlier.
i have added a checks to verify whether the container still exists before trying to stop it.

Vaishnav88sk (changes requested):

pkg/config/localconfig_test.go
Syedowais312: [060992b](https://github.com/microcks/microcks-cli/commit/060992b1eee67a9e36e7b90cc3371e445012e773) added

cmd/stop.go
```diff
@@ -40,6 +40,14 @@ func NewStopCommand(globalClientOpts *connectors.ClientOptions) *cobra.Command {
 			errors.CheckError(err)
 			defer containerClient.CloseClient()
 
+			exists, _ := containerClient.ContainerExists(instance.ContainerID)
+			errors.CheckError(err)
```
Vaishnav88sk: There is a bug here: `ContainerExists(...)` returns `(bool, error)`, but the error is being discarded with `_`, and then `errors.CheckError(err)` is checking the previous `err` value instead.

Please capture the returned error here and validate that one.

Vaishnav88sk: Also line ~70, 
This should stay as `RemoveInstance(instance.ContainerID)`, not `RemoveInstance(instance.Name)`.

Switching it back to `instance.Name` reintroduces the original issue this PR is fixing, since `RemoveInstance()` was changed to match by `ContainerID`.

Vaishnav88sk (changes requested):
Please resolve that. After that LGTM. 👍🏻

cmd/stop.go
Syedowais312: I have made the require changes @Vaishnav88sk

Caesarsage:
Thanks @Syedowais312 

The reconcile logic on master resets the stale entry in place (Status/ContainerID cleared, then reused) rather than appending a new record — see the "Reconcile before trusting it" block in cmd/start.go.

Could you re-verify duplicates still accumulate on a current build? If not, this can be closed.

Syedowais312:
> Thanks @Syedowais312
> 
> The reconcile logic on master resets the stale entry in place (Status/ContainerID cleared, then reused) rather than appending a new record — see the "Reconcile before trusting it" block in cmd/start.go.
> 
> Could you re-verify duplicates still accumulate on a current build? If not, this can be closed.

Hey @Caesarsage, I pulled the latest `master` and verified again. The bug still seems to exist.

 The reason is around this line:

  https://github.com/microcks/microcks-cli/blob/0dd8f55ba4aa67b3c6e9918ad29bd5b3d6b9dbc5/cmd/start.go#L102

  Here we clear `instance.Status` and `instance.ContainerID` when the old container no longer exists. Because the status becomes empty, the switch falls through to the  `default` case and creates a new container.

  But the old instance entry is still present in the config with the old container ID. Later, `UpsertInstance` compares using `ContainerID`, so the recreated instance has a new ID and gets appended instead of replacing the old entry.

  In this PR, I removed the stale instance before recreating it. I also changed `RemoveInstance` to remove by `ContainerID`, because that is the exact stale entry we are trying to clean up.

Syedowais312:
Hey, I've rebased the branch onto the latest master. Could you take a look when you get a chance?

cmd/stop.go
```diff
@@ -62,6 +62,16 @@ func NewStopCommand(globalClientOpts *connectors.ClientOptions) *cobra.Command {
...
+			}
+			if !exists {
+				fmt.Printf("Container for instance %s no longer exists\n", instance.Name)
+				fmt.Printf("Run 'microcks start --name %s' to bring the container back\n", instance.Name)
+				return nil
+			}
+
 			if err := containerClient.StopContainer(instance.ContainerID); err != nil {
 				return errors.Wrap(errors.KindEnvironment, fmt.Errorf("failed to stop container: %w", err))
 			}
```
Vaishnav88sk: Returning `nil` early here prevents the CLI from crashing (which is a great improvement over the old behavior where `StopContainer` would throw an error), but it completely bypasses the configuration cleanup that happens at the bottom of the file.
**Suggestion**: Instead of returning early, we should just skip the `StopContainer` call and proceed directly to the `// update configs` section so the local state accurately reflects that the container is gone.

```suggestion
			if !exists {
				fmt.Printf("Container for instance %s no longer exists\n", instance.Name)
			} else {
				if err := containerClient.StopContainer(instance.ContainerID); err != nil {
					return errors.Wrap(errors.KindEnvironment, fmt.Errorf("failed to stop container: %w", err))
				}
				fmt.Println("")
				log.Printf("Instance %s stopped successfully", instance.Name)
			}
```

Vaishnav88sk (changes requested):

Issue #473, opened 2026-06-23
CI/CD Pipeline False Positives: `import-dir` Exits with Code 0 on Failures


aniket866:
### Description
The `import-dir` command scans a directory and attempts to import all matching files. If one or more files fail to import, it records the failure in `ImportResult.FailedCount` and prints an error message. However, the command runner does not check the failed count to return a non-zero exit code.

### Impact
In CI/CD environments (GitHub Actions, Tekton, GitLab CI), the CLI exits with status `0`, indicating success even if all API specifications failed to import. This hides deployment failures.

### Code Reference
* [cmd/importDir.go](file:///c:/Users/Hp/microcks-cli/cmd/importDir.go#L162-L200)

### Diagram
```mermaid
graph TD
    A[Start import-dir] --> B[Scan Directory]
    B --> C[Import File 1: Success]
    B --> D[Import File 2: FAILED]
    C & D --> E[Display Results Summary]
    E --> F["Exit Status Code 0 (Success)"]
    F --> G["CI/CD pipeline reports green/success despite failures"]
```

### Steps to Reproduce
1. Run `microcks import-dir ./specs` where at least one spec has invalid content.
2. The console displays `✗ Failed: ...`.
3. Check the exit code of the process (`echo $?` or `$LASTEXITCODE` in PowerShell). It returns `0`.

---

Caesarsage:
Thank you @aniket866 

This has been Fixed on master, import-dir now returns a non-zero exit when any file fails: importDirectoryPartialFailure wraps FailedCount > 0 into a KindAPI error that the exit funnel maps to a failure code. 

I believe this can be closed.

Issue #446, opened 2026-05-26
test command help text does not document required positional arguments


hazelmayank:
## Description

`microcks test --help` advertises the command's flags, but does not document
the three required positional arguments. A new user reading `--help` has no
indication that the command needs `<apiName:apiVersion> <testEndpoint> <runner>`.

### Reproduction

```
$ microcks test --help
Run tests on Microcks

Usage:
  microcks test [flags]

Flags:
      --filteredOperations string   List of operations to launch a test for
      ...
```

### Current behavior on missing args

The required-args contract is only discoverable by running the command with
no args and reading the error message printed by the manual check at
`cmd/test.go:51`:

```
test command require <apiName:apiVersion> <testEndpoint> <runner> args
```

This is a UX gap — the help text should be self-sufficient. It also means
the docs generated under `documentation/cmd/test.md` lack the same info.

## Proposed fix

Two changes in `cmd/test.go`:

1. **Document the args in Cobra metadata.**
   - Update `Use:` to `test <apiName:apiVersion> <testEndpoint> <runner>`.
   - Expand `Long:` to describe each required argument.
   - Add an `Example:` block showing a realistic invocation.

2. **Use Cobra's declarative arg validator.**
   - Set `Args: cobra.ExactArgs(3)` on the command. Cobra rejects calls with
     the wrong number of positional args automatically and prints the usage
     hint, giving better UX than the current manual check.
   - Remove the now-redundant `if len(os.Args) < 4 { ... }` block in `Run`.

No change to flags, server interaction, or exit-code semantics. The
per-argument checks for empty / dash-prefixed values are kept for defense.

## Environment

- Branch: `master` (development version `1.0.3`)
- File: `cmd/test.go`

Caesarsage:
Thank you @hazelmayank 


Fixed on master, Use: now reads "test <apiName:apiVersion> <testEndpoint> <runner>" and the command declares cobra.ExactArgs(3), so --help shows the full contract and a wrong arg count produces a proper usage error. 

I believe this can be closed.

Vaishnav88sk:
Confirmed on master: `cmd/test.go` now sets `Use: "test <apiName:apiVersion> <testEndpoint> <runner>"` and `Args: cobra.ExactArgs(3)`, so `--help` shows the full contract and wrong arg counts fail with a usage error. Closing.
@Harsh4902

PR #482, opened 2026-06-24
fix(watcher): skip context when Microcks client cannot be created

### LINKED ISSUE #475: Nil Pointer Dereference in Watcher Executor -> dereference panic.
### Diagram
```mermaid
sequenceDiagram
    participant Watcher
    participant Executor
    participant Client as connectors.NewClient
    Watcher->>Executor: TriggerImport(entry)
    Executor->>Client: NewClient(*globalClientOpts)
    Note over Client: Fails (e.g. invalid context)
    Client-->>Executor: nil, err
    Note over Executor: Logs error, but DOES NOT return!
    Executor->>Executor: mc.UploadArtifact(entry.FilePath, entry.MainArtifact)
    Note over Executor: Panic: nil pointer dereference (mc is nil)
```

### Description
In `pkg/watcher/executor.go`, the `TriggerImport` function instantiates the Microcks client using `connectors.NewClient`. If this fails (e.g., due to an invalid context or server address), it prints an error message but continues execution instead of returning early. It subsequently calls `mc.UploadArtifact()`, causing a nil pointer dereference panic.

### Impact
The watcher daemon crashes with a stack trace instead of handling the connection/context resolution failure gracefully.

### Steps to Reproduce
1. Configure a watch entry with an invalid or expired context.
2. Modify the watched file to trigger an fsnotify event.
3. The watch manager invokes `TriggerImport`, which prints `[ERROR] Cannot connect to Microcks client...` and immediately panics with a segmentation fault.

---

SAY-5:
## Problem

In `TriggerImport`, when `connectors.NewClient` fails it logs the error but
keeps going, so the code reaches `mc.UploadArtifact` with a nil client and the
watcher process panics with a nil pointer dereference (#475). This happens for
example when a watch entry references a context that cannot be resolved.

## Fix

`continue` to the next context after logging the connection error, so a failed
context is skipped instead of crashing the daemon.

Added a regression test in `pkg/watcher` that drives `TriggerImport` with an
unresolvable context; it panics without the change and passes with it.

Fixes #475

### DIFF
--- a/pkg/watcher/executor.go
+++ b/pkg/watcher/executor.go
@@ -33,6 +33,7 @@ func TriggerImport(entry config.WatchEntry) {
 			mc, err = connectors.NewClient(*globalClientOpts)
 			if err != nil {
 				fmt.Printf("[ERROR] Cannot connect to Microcks client: %v in context '%s'\n", err, context)
+				continue
 			}
 		} else {
 			// We have no config file, so just create a client with context as server URL.
--- a/pkg/watcher/executor_test.go
+++ b/pkg/watcher/executor_test.go
@@ -1,0 +1,37 @@
+package watcher
+
+import (
+	"os"
+	"path"
+	"testing"
+
+	"github.com/microcks/microcks-cli/pkg/config"
+)
+
+// TestTriggerImportSkipsContextWhenClientFails ensures TriggerImport does not
+// panic when the Microcks client cannot be created for a context (for example
+// an unknown or unresolvable context). Before the fix it dereferenced a nil
+// client and crashed the watcher process.
+func TestTriggerImportSkipsContextWhenClientFails(t *testing.T) {
+	home := t.TempDir()
+	t.Setenv("HOME", home)
+
+	// Write an empty local config so that resolving any context fails and
+	// NewClient returns a nil client with an error.
+	cfgDir := path.Join(home, ".config", "microcks")
+	if err := os.MkdirAll(cfgDir, 0o755); err != nil {
+		t.Fatalf("failed to create config dir: %v", err)
+	}
+	if err := os.WriteFile(path.Join(cfgDir, "config"), []byte("{}"), 0o600); err != nil {
+		t.Fatalf("failed to write config: %v", err)
+	}
+
+	entry := config.WatchEntry{
+		FilePath:     path.Join(home, "spec.json"),
+		Context:      []string{"missing-context"},
+		MainArtifact: true,
+	}
+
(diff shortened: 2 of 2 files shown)


Caesarsage:
Thank you @aniket866 

This issue has been fixed on master — TriggerImport now continues to the next context when connectors.NewClient fails, so the nil client can never reach UploadArtifact. 

I believe this can be closed.

Vaishnav88sk:
Same comment: https://github.com/microcks/microcks-cli/pull/496#issuecomment-5312198162

Caesarsage:
You're absolutely right, @Vaishnav88sk, thanks for checking, and apologies for the noise. I re-read the full control flow and my comment was wrong: the continue only exists in the else branch (no-config path via NewMicrocksClient). The config-path branch logs the NewClient error and falls through to mc.UploadArtifact on a nil client — exactly the #475 scenario. So #475 is still valid and this PR still applies.

Vaishnav88sk:
> You're absolutely right, @Vaishnav88sk, thanks for checking, and apologies for the noise. I re-read the full control flow and my comment was wrong: the continue only exists in the else branch (no-config path via NewMicrocksClient). The config-path branch logs the NewClient error and falls through to mc.UploadArtifact on a nil client — exactly the #475 scenario. So #475 is still valid and this PR still applies.

Thanks for the clarification. Could you please recheck the comments yu made previously on issues/PRs about the closure. I think there are some that not yet solved or needed.

Caesarsage:
> > You're absolutely right, @Vaishnav88sk, thanks for checking, and apologies for the noise. I re-read the full control flow and my comment was wrong: the continue only exists in the else branch (no-config path via NewMicrocksClient). The config-path branch logs the NewClient error and falls through to mc.UploadArtifact on a nil client — exactly the #475 scenario. So #475 is still valid and this PR still applies.
> 
> Thanks for the clarification. Could you please recheck the comments yu made previously on issues/PRs about the closure. I think there are some that not yet solved or needed.

Done — I went back over every comment I posted, re-verifying each against master with full code reads this time. 

Everything held up so far: importDirectoryPartialFailure for #473/#379, reference-counting deleteContext for #421/#427, ResolveContext in logout for #419, waitForReady in start for #303/#311, ExactArgs + --output in test for #446/#332, resp.Body + regression test for #337/#375, and zero panic( sites left in pkg/connectors for the #489 family.

One refinement rather than a correction: the "client_secret gap" I noted on #265/#333 turned out to be bigger — it's now properly documented in #503 (JSON bodies bypass redaction), with a fix in review at #504.

I am still reviewing pending ones which I will continue later

PR #371, opened 2026-05-12
feat: add global output formatter support (text/json/yaml)

cotishq:
### Description

- Added a global `--output` flag to `microcks test` with supported values: `text` (default), `json`, and `yaml`.
- Introduced a reusable output layer in `pkg/output` with shared formatter routing and dedicated formatters (`TextFormatter, JSONFormatter, YAMLFormatter`) so other commands can reuse the same pattern later.
- Added `TestResultJSON` mapping structs that mirror `connectors.TestResult` fields (including nested test case/step fields) and used them for structured JSON/YAML serialization.
- Updated test command execution flow to route by --output format while preserving existing default text behavior and existing exit code semantics.
- Ensured structured outputs `(json/yaml)` are emitted to `stdout` for automation use-cases, with progress/info logs routed to `stderr`.
- Added/updated tests in pkg/output for JSON/YAML formatting and formatter factory selection.
- Updated command docs in `documentation/cmd/test.md` to include the new `--output` flag and supported values.

### Related issue(s)

Resolves #301

PR #488, opened 2026-07-01
feat(test): add --output (text/json/yaml/github-actions) for machine-readable results

Caesarsage:
Adds `--output` to `microcks test`, with four formats:

| Value | Output |
|-------|--------|
| `text` (default) | Human-readable summary (unchanged default behaviour) |
| `json` | Full `TestResult` as JSON |
| `yaml` | Full `TestResult` as YAML |
| `github-actions` | GitHub Actions annotations + log groups + job step summary |

Works for both the normal server path and `--dry-run`, since it's wired into the
shared test engine (`runTestAndWait`).

## github-actions format

- `::group::` / `::endgroup::` log-folding per operation
- `::error::` per failing step, with the validation message
- `::error file=<spec>,line=<n>::` when the local spec is known (`--dry-run --artifact`),
  mapping the failing operation to its line so the annotation lands on the diff
- `::notice::` for passing operations, gated by `MICROCKS_ACTIONS_VERBOSE`
- A per-operation markdown table appended to `$GITHUB_STEP_SUMMARY`

## Tests

- Unit tests for all four formatters, the github-actions escaping, the step
  summary, and the OpenAPI line-mapping (`pkg/output`).

## Notes

- **Supersedes #371** (same `--output`, extended with github-actions + diff
  annotations) — that PR can be closed.
- A runnable end-to-end demo workflow lives on the https://github.com/Caesarsage/microcks-cli/actions/runs/28505381116
Refs #255

### DIFF
--- a/pkg/connectors/microcks_client.go
+++ b/pkg/connectors/microcks_client.go
@@ -50,6 +50,7 @@ type MicrocksClient interface {
 	SetOAuthToken(oauthToken string)
 	CreateTestResult(serviceID string, testEndpoint string, runnerType string, secretName string, timeout int64, filteredOperations string, operationsHeaders string, oAuth2Context string) (string, error)
 	GetTestResult(testResultID string) (*TestResultSummary, error)
+	GetFullTestResult(testResultID string) (*TestResult, error)
 	UploadArtifact(specificationFilePath string, mainArtifact bool) (string, error)
 	DownloadArtifact(artifactURL string, mainArtifact bool, secret string) (string, error)
 }
@@ -65,6 +66,37 @@ type TestResultSummary struct {
 	ElapsedTime    int32  `json:"elapsedTime"`
 	Success        bool   `json:"success"`
 	InProgress     bool   `json:"inProgress"`
+}
+
+// TestResult represents a full Microcks TestResult including per-operation detail.
+type TestResult struct {
+	ID              string           `json:"id"`
+	Version         int32            `json:"version"`
+	TestNumber      int32            `json:"testNumber"`
+	TestDate        int64            `json:"testDate"`
+	TestedEndpoint  string           `json:"testedEndpoint"`
+	ServiceID       string           `json:"serviceId"`
+	ElapsedTime     int32            `json:"elapsedTime"`
+	Success         bool             `json:"success"`
+	InProgress      bool             `json:"inProgress"`
+	TestCaseResults []TestCaseResult `json:"testCaseResults"`
+}
+
+// TestCaseResult is the result for a single operation within a TestResult.
+type TestCaseResult struct {
+	Success         bool             `json:"success"`
+	ElapsedTime     int32            `json:"elapsedTime"`
+	OperationName   string           `json:"operationName"`
+	TestStepResults []TestStepResult `json:"testStepResults"`
+}
+
+// TestStepResult is the result for a single request/message within a TestCaseResult.
+type TestStepResult struct {
+	Success          bool   `json:"success"`
+	ElapsedTime      int32  `json:"elapsedTime"`
+	RequestName      string `json:"requestName"`
+	EventMessageName string `json:"eventMessageName"`
+	Message          string `json:"message"`
 }
 
 // HeaderDTO represents an operation header passed for Test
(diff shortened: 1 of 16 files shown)


pkg/output/yaml_formatter.go
```diff
@@ -0,0 +1,43 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package output
+
+import (
+	"encoding/json"
+
+	"github.com/microcks/microcks-cli/pkg/connectors"
+	"gopkg.in/yaml.v2"
```
Vaishnav88sk, line 22: In `openapi_linemap` file, you used `yaml.v3`, so why `v2` here?

pkg/output/openapi_linemap.go
```diff
@@ -0,0 +1,82 @@
...
+	"strings"
+
+	yamlv3 "gopkg.in/yaml.v3"
+)
+
+// operationLine returns the 1-based line of an operation (e.g. "GET /products")
+// within an OpenAPI spec file, or 0 if it can't be determined (non-YAML spec,
+// parse error, or operation not found). yaml.v3 nodes carry line numbers, which
+// yaml.v2 does not expose.
+func operationLine(specPath, operationName string) int {
```
Vaishnav88sk, line 29: Nit: the function name `operationLine` and the `paths` key lookup hard-code an assumption that this is an OpenAPI spec. But `microcks test --dry-run` also works with AsyncAPI, Postman Collections, and GraphQL artifacts.

For AsyncAPI, the top-level key is `channels` (AsyncAPI 2.x) or `channels`/`operations` (AsyncAPI 3.x), not `paths`. Postman Collections are JSON, not YAML at all.

pkg/output/openapi_linemap.go
```diff
@@ -0,0 +1,82 @@
...
+
+	_, pathsNode := mappingEntry(root, "paths")
+	if pathsNode == nil {
+		return 0
+	}
+	_, pathNode := mappingEntry(pathsNode, path)
+	if pathNode == nil {
+		return 0
+	}
+	line, _ := mappingEntry(pathNode, method)
```
Vaishnav88sk, line 57: The blank identifier `_` here silently discards whether the method lookup found anything. If `mappingEntry` returns `(0, nil)` for a missing key, that's already handled correctly since you return `0`.

But would it be worth extracting a named variable for clarity?

```go
line, methodNode := mappingEntry(pathNode, method)
if methodNode == nil {
    return 0
}
return line

File: pkg/output/openapi_linemap.go [hunk 4]
```diff
@@ -0,0 +1,82 @@
...
+	return line
+}
+
+// splitOperation parses "GET /products" into ("get", "/products", true).
+func splitOperation(operationName string) (method, path string, ok bool) {
```
Vaishnav88sk, line 62: This works great for the OpenAPI `METHOD /path` convention, but Microcks also supports AsyncAPI (e.g. `SUBSCRIBE channel`) and GraphQL (e.g. `getPastries`) where the split won't match, though the safe `false` fallback handles that correctly.

Worth adding a short doc comment to make this intentional scope explicit.

Caesarsage:
/hold

Working on a structure logger and helper function from last mentorship call

Caesarsage:
if #489 or #490 or #491 gets merge, I will rework this PR to use the new error format

Harsh4902 (dismissed):
New dependancies are in place. GTG

Vaishnav88sk:
I think the review comments are still not resolved. After that 👍🏻

Caesarsage:
> I think the review comments are still not resolved. After that 👍🏻

Yes, once the other PRs are merged, I will rebase and rework attending to your reviews. this is so as to avoid many conflicts. As stated here.

https://github.com/microcks/microcks-cli/pull/488#issuecomment-5048696050

pkg/connectors/microcks_client.go
```diff
@@ -468,6 +500,43 @@ func (c *microcksClient) GetTestResult(testResultID string) (*TestResultSummary,
...
+	}
+	defer resp.Body.Close()
+
+	config.DumpResponseIfRequired("Microcks for getting full test result", resp, true)
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, err
+	}
+
```
Vaishnav88sk, line 538: This method doesn’t check `resp.StatusCode` before unmarshalling, unlike the other client methods (e.g. `CreateTestResult` checks `201`). If the API returns a non-2xx with an error body, this will either unmarshal into an empty `TestResult` or fail with a confusing parse error.

Please add a status check here, similar to the rest of the client:

```suggestion
if resp.StatusCode != http.StatusOK {
    return nil, errors.Wrapf(errors.KindAPI, "Microcks returned HTTP %d: %s", resp.StatusCode, strings.TrimSpace(string(body)))
}
```

Vaishnav88sk (dismissed):
Last fix please!! 🙂

Caesarsage:
Thank you for the review. I have updated it @Vaishnav88sk

Vaishnav88sk:
LGTM @Harsh4902

Vaishnav88sk:
Thanks for the contribution!! @Caesarsage

Harsh4902 (approved):
Good to go. Thanks @Caesarsage and @Vaishnav88sk

PR #481, opened 2026-06-23
Uncontrolled Panics in API Client on Unexpected Server Responses

aniket866:
### Issue
In `pkg/connectors/microcks_client.go`, functions like `GetKeycloakURL` and `GetTestResult` use `panic(err)` or `panic(err.Error())` if response body reading or JSON deserialization fails.

### Impact
If a reverse proxy, API gateway, or cloud load balancer returns an HTML error page (e.g., 502 Bad Gateway, 504 Gateway Timeout, 403 Forbidden) instead of valid JSON, the CLI crashes with a Go stack trace instead of presenting a clean error message to the user.

### Code Reference
* [pkg/connectors/microcks_client.go](file:///c:/Users/Hp/microcks-cli/pkg/connectors/microcks_client.go#L241-L248)
* [pkg/connectors/microcks_client.go](file:///c:/Users/Hp/microcks-cli/pkg/connectors/microcks_client.go#L428)

### Diagram
```mermaid
sequenceDiagram
    participant CLI
    participant Proxy as Reverse Proxy (Nginx/Cloudflare)
    participant Server as Microcks Server
    CLI->>Proxy: GET /api/keycloak/config
    Proxy--xServer: Connection Refused / Timeout
    Proxy-->>CLI: HTTP 504 Gateway Timeout (HTML body)
    Note over CLI: io.ReadAll() succeeds but JSON parsing fails
    Note over CLI: json.Unmarshal(body, &configResp) errors out
    CLI->>CLI: panic(err) (Hard crash with stack trace)
```

### Steps to Reproduce
1. Target a server address that serves non-JSON content (e.g., any public landing page returning HTML).
2. Run `microcks login <URL>`.
3. The client fails to unmarshal the HTML body and crashes with `panic: invalid character '<' looking for beginning of value`.

---
Closes #478

PR #489, opened 2026-07-09
fix(cli): classify failures and partial de-panic in the client library

Caesarsage:
Part of the Microcks CLI v2 work https://github.com/microcks/microcks-cli/issues/255.

Introduces a small **Failure Kind** vocabulary in `pkg/errors` (`Kind`, `Wrap`/`Wrapf`, `KindOf`, and an `ErrTestFailed` sentinel) and makes the client library **return classified errors instead of panicking or exiting**.

### Why
`pkg/connectors` and `keycloak_client` previously `panic`'d on runtime errors (bad response body, unreachable server). A library that can take down its host process is unsafe to embed, which the planned VS Code extension needs to do. Kinds are the library's vocabulary; exit codes stay a CLI concern (added in next PR ()), so `pkg/*` never depends on them.

### Notable
Fixes two real latent bugs found while de-panicking:
- Keycloak swallowed request-creation errors: nil-pointer dereference.
- Unchecked `openIDResp["access_token"].(string)`:  **the CLI crashed on a Keycloak 401** instead of reporting it.

### Scope / safety
Additive and **non-breaking**: the legacy `CheckError`/`Fatal` shim and the existing constructor signatures are untouched here (removed/changed in later PRs). Build + tests green.

### DIFF
--- a/pkg/connectors/keycloak_client.go
+++ b/pkg/connectors/keycloak_client.go
@@ -26,6 +26,7 @@ import (
 	"strings"
 
 	"github.com/microcks/microcks-cli/pkg/config"
+	"github.com/microcks/microcks-cli/pkg/errors"
 	"golang.org/x/oauth2"
 )
 
@@ -45,12 +46,12 @@ type keycloakClient struct {
 }
 
 // NewKeycloakClient build a new KeycloakClient implementation
-func NewKeycloakClient(realmURL string, username string, password string) KeycloakClient {
+func NewKeycloakClient(realmURL string, username string, password string) (KeycloakClient, error) {
 	kc := keycloakClient{}
 
 	u, err := url.Parse(realmURL)
 	if err != nil {
-		panic(err)
+		return nil, errors.Wrap(errors.KindUsage, fmt.Errorf("invalid Keycloak URL %q: %w", realmURL, err))
 	}
 	kc.BaseURL = u
 	kc.Username = username
@@ -65,7 +66,7 @@ func NewKeycloakClient(realmURL string, username string, password string) Keyclo
 	} else {
 		kc.httpClient = http.DefaultClient
 	}
-	return &kc
+	return &kc, nil
 }
 
 // ConnectAndGetToken implementation on keycloakClient structure
@@ -88,7 +89,7 @@ func (c *keycloakClient) ConnectAndGetToken() (string, error) {
 
 	resp, err := c.httpClient.Do(req)
 	if err != nil {
-		return "", err
+		return "", errors.Wrap(errors.KindConnection, err)
 	}
 	defer resp.Body.Close()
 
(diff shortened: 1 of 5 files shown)


pkg/errors/error_test.go
Vaishnav88sk, line 1: Please add copyright block at start

pkg/connectors/microcks_client.go
```diff
@@ -194,6 +194,8 @@ func NewMicrocksClient(apiURL string) MicrocksClient {
 
 	u, err := url.Parse(apiURL)
 	if err != nil {
+		// url.Parse only fails on a malformed URL; returning it needs a
+		// signature change, done with the RunE command migration.
 		panic(err)
 	}
```
Vaishnav88sk, line 200: This reintroduces a panic in library code, which conflicts with the stated goal of this PR.

Even if malformed URLs are a narrower case, a library constructor still should not panic here. If we cannot change the constructor signature in this PR, then I’d rather keep this work scoped differently than claim we have removed panics from the client layer.

pkg/connectors/keycloak_client.go
```diff
@@ -50,6 +51,8 @@ func NewKeycloakClient(realmURL string, username string, password string) Keyclo
 
 	u, err := url.Parse(realmURL)
 	if err != nil {
+		// url.Parse only fails on a malformed URL; returning it needs a
+		// signature change, done with the RunE command migration.
 		panic(err)
```
Vaishnav88sk, line 56: Same concern here: this keeps a panic path in the client library.

For embeddability, constructors should ideally return errors instead of terminating the host process. If that signature migration has to happen separately, please call that out more explicitly and avoid over-claiming the scope of this PR.

pkg/connectors/microcks_client.go
```diff
@@ -239,24 +241,29 @@ func (c *microcksClient) GetKeycloakURL() (string, error) {
...
 
-	// Retrieve auth server url and realm name.
-	enabled := configResp["enabled"].(bool)
-	authServerURL := configResp["auth-server-url"].(string)
-	realmName := configResp["realm"].(string)
+	// Return 'null' if Keycloak is disabled.
+	if enabled, _ := configResp["enabled"].(bool); !enabled {
+		return "null", nil
+	}
```
Vaishnav88sk, line 259: We should not silently treat a missing / malformed `enabled` field as “Keycloak disabled”.

If the response shape is wrong, please return an API error instead. Otherwise a bad server/proxy response can be misclassified as a valid “no Keycloak” setup.

Vaishnav88sk:
supersede #481

pkg/connectors/microcks_client.go
Caesarsage: Thanks, this has been fixed on the next PR

https://github.com/microcks/microcks-cli/pull/490/changes#diff-bac918769b81eb6fc95691c8e3f0c6131680f522f5b3e4dd1a53be291ffabbdaL52

pkg/connectors/keycloak_client.go
Caesarsage: Thanks, this has been fixed on the next PR

https://github.com/microcks/microcks-cli/pull/490

pkg/connectors/microcks_client.go
Caesarsage: Similarly, it was addressed on the #490 PR

pkg/errors/error_test.go
Caesarsage: added

Vaishnav88sk:
DCO is required.

Caesarsage:
fixed @Vaishnav88sk

Vaishnav88sk:
Suggestion: 
- Update the PR title/description to explicitly scope this PR as a partial de-panic, with the remaining panic paths and the enabled-field issue called out as known follow-ups landing in #490

Vaishnav88sk:
After that LGTM. 👍🏻

Harsh4902 (approved):
LGTM

PR #490, opened 2026-07-09
refactor(cli): migrate commands to RunE with a single exit funnel

Caesarsage:
depend on https://github.com/microcks/microcks-cli/pull/489

Part of the Microcks CLI v2 work (https://github.com/microcks/microcks-cli/issues/255). Stacked on PR https://github.com/microcks/microcks-cli/pull/489.

- All ten commands move from `Run` + scattered `os.Exit`/`log.Fatal` to **`RunE` returning errors**.
- A single funnel — `cmd.Handle`, called only by `main` — prints to stderr and maps Failure Kind → exit code (`cmd/exit.go`). Nothing else exits.
- A non-conforming contract test travels as the silent `ErrTestFailed` sentinel → exit `1` with no spurious error line (mirrors `kubectl diff`).

No command/flag interface changes; failure output and exit codes intentionally change.

### Exit codes (new, user-visible)
`0` = success / non-zero = failure still holds, and `1` still means "contract test failed". Operational failures now carry distinct codes so CI can branch on *why*:

| Code | Meaning |
| ---- | ------- |
| 0 | success / contract conformed |
| 1 | contract test failed (a clean run) |
| 2 | usage — bad arguments or flags |
| 11 | connection — couldn't reach Microcks/Keycloak |
| 12 | API — server rejected the request / unusable response |
| 13 | not-found — remote resource missing |
| 14 | environment — local precondition (container runtime, image, readiness) |
| 20 | generic — unclassified failure |

These codes were previously undocumented and inconsistent (`1`/`2`/`20` interchangeably), so there is no prior contract to break. Documented in `README.md` + `documentation/error-handling.md`.


Caesarsage:
HOLD 

till PR https://github.com/microcks/microcks-cli/pull/489 is review and merged

Vaishnav88sk:
Stacked PR of #489

cmd/testDryRun.go
```diff
@@ -218,24 +220,21 @@ func terminateContainer(container *microcks.MicrocksContainer) {
...
 	if err != nil {
-		fmt.Printf("Failed to resolve artifact path: %s\n", err)
-		return false
+		return fmt.Errorf("failed to resolve artifact path: %w", err)
 	}
 	if err := watcher.Add(filepath.Dir(artifactPath)); err != nil {
-		fmt.Printf("Failed to watch %s: %s\n", filepath.Dir(artifactPath), err)
-		return false
+		return fmt.Errorf("failed to watch %s: %w", filepath.Dir(artifactPath), err)
 	}
```
Vaishnav88sk, line 238: These new error paths are still unclassified `fmt.Errorf(...)` returns.

Since this PR is introducing a kind-based exit funnel, these should be wrapped consistently as well (likely `KindEnvironment` for watcher/runtime setup problems, or `KindUsage` where appropriate). Otherwise these cases still collapse to the generic `20` path.

cmd/import.go
```diff
@@ -36,12 +35,10 @@ func NewImportCommand(globalClientOpts *connectors.ClientOptions) *cobra.Command
...
 		Args:  cobra.MaximumNArgs(1),
-		Run: func(cmd *cobra.Command, args []string) {
+		RunE: func(cmd *cobra.Command, args []string) error {
 			// Parse subcommand args first.
 			if len(args) == 0 {
-				fmt.Println("import command require <specificationFile1[:primary],specificationFile2[:primary]> args")
-				cmd.HelpFunc()(cmd, args)
-				os.Exit(1)
+				return errors.Wrapf(errors.KindUsage, "import requires a <specificationFile1[:primary],specificationFile2[:primary]> argument")
 			}
```
Vaishnav88sk, line 42: One behavioral change to call out here: this no longer prints help/usage, whereas the previous implementation did.

I’m fine with the `RunE` refactor, but if the goal is “no impact on the binary”, we should either preserve the old help/usage behavior for bad invocations or explicitly document/test that this user-facing behavior is changing.

cmd/login.go
```diff
@@ -57,26 +57,27 @@ microcks login http://localhost:8080 --sso --sso-port
...
+		RunE: func(cmd *cobra.Command, args []string) error {
 			ctx := cmd.Context()
 			var server string
 
 			//Check if server name is provided or not
 			if len(args) != 1 {
-				cmd.HelpFunc()(cmd, args)
-				os.Exit(1)
+				return errors.Wrapf(errors.KindUsage, "login requires exactly one SERVER argument")
 			}
```
Vaishnav88sk, line 67: Same comment here: this changes the UX on invalid invocation from “show help + exit” to “print one error line + exit”.

That may be acceptable, but it is user-visible behavior and contradicts the current PR description of “no impact on the binary”. Please either preserve the old behavior or call out the change explicitly.

cmd/importDir.go
```diff
@@ -197,6 +191,7 @@ func NewImportDirCommand(globalClientOpts *connectors.ClientOptions) *cobra.Comm
 			}
 
 			fmt.Printf("\nImport completed: %d/%d files imported successfully\n", result.SuccessCount, result.TotalFiles)
```
Vaishnav88sk, line 193: Please double-check this path against the intended `import-dir` semantics.

As written, partial failures still return `nil`, so the single exit funnel will still exit `0` even when some files failed. If we want `import-dir` to be CI-friendly, this likely needs to return a non-zero error when `FailedCount > 0`.

pkg/connectors/keycloak_client.go
```diff
@@ -45,12 +46,12 @@ type keycloakClient struct {
...
 // NewKeycloakClient build a new KeycloakClient implementation
-func NewKeycloakClient(realmURL string, username string, password string) KeycloakClient {
+func NewKeycloakClient(realmURL string, username string, password string) (KeycloakClient, error) {
 	kc := keycloakClient{}
 
 	u, err := url.Parse(realmURL)
 	if err != nil {
-		panic(err)
+		return nil, errors.Wrap(errors.KindUsage, fmt.Errorf("invalid Keycloak URL %q: %w", realmURL, err))
 	}
```
Vaishnav88sk, line 55: I flagged in #489 and you said was fixed here. Please point to the current line number and confirm it now returns errors.Wrap(errors.KindUsage or KindAPI, err) instead of panic(err).

pkg/connectors/microcks_client.go
```diff
@@ -194,7 +194,7 @@ func NewMicrocksClient(apiURL string) MicrocksClient {
 
 	u, err := url.Parse(apiURL)
 	if err != nil {
-		panic(err)
+		return nil, errors.Wrap(errors.KindUsage, fmt.Errorf("invalid server URL %q: %w", apiURL, err))
 	}
 	mc.APIURL = u
```
Vaishnav88sk, line 200: Same as `keycloak_client.go`, confirm the url.Parse panic here is replaced with a returned classified error, and share the line number.

Vaishnav88sk (changes requested):
Please resolve the conflicts.

Also add copyright block for `pkg/errors/error_test.go`

pkg/connectors/keycloak_client.go
Caesarsage: this is returning the errors helper as i mentioned on #489. I am not sure I understand the comment here

pkg/connectors/microcks_client.go
Caesarsage: this is returning the errors helper as i mentioned on #489 and all are documented on the documentation folder. I am not sure I understand the comment here

cmd/import.go
Caesarsage: Thanks for catching this. I will make a fix for this by creating am helper usage function

cmd/testExecutor.go
Vaishnav88sk, line 1: These are still plain `fmt.Errorf(...)` returns.

Since this PR is introducing a kind-based exit funnel, these paths should also be classified appropriately instead of falling back to generic exit code `20`.

Vaishnav88sk (dismissed):

Vaishnav88sk:
Some need to check:
- usage/help preservation is not applied consistently yet; some commands still return plain `KindUsage` errors instead of using the new `usageErrorf(...)` path
- the PR description still says “No impact on the binary”, which is too strong given the user-visible exit-code/error-handling changes

Please clean up those last consistency issues and then I’m happy to re-check.

Caesarsage:
> Some need to check:
> 
> * usage/help preservation is not applied consistently yet; some commands still return plain `KindUsage` errors instead of using the new `usageErrorf(...)` path
> * the PR description still says “No impact on the binary”, which is too strong given the user-visible exit-code/error-handling changes
> 
> Please clean up those last consistency issues and then I’m happy to re-check.

Thanks for the review. The usage/help preservation was only added on paths that already printed usage before this PR. So I wouldn’t call the remaining commands a consistency regression in this PR, since they did not previously have that behavior.
I’m happy to raise a follow-up PR to standardize usage/help output across all commands. For this PR, my intent was narrower: migrate commands from Run to RunE and route returned errors through the new centralized handler.

I agree the PR description should not say “No impact on the binary” so strongly, since the exit-code/error-handling behavior is user-visible. I’ll reword that part.

Vaishnav88sk:
> Thanks for the review. The usage/help preservation was only added on paths that already printed usage before this PR. So I wouldn’t call the remaining commands a consistency regression in this PR, since they did not previously have that behavior. I’m happy to raise a follow-up PR to standardize usage/help output across all commands. For this PR, my intent was narrower: migrate commands from Run to RunE and route returned errors through the new centralized handler.
> 
> I agree the PR description should not say “No impact on the binary” so strongly, since the exit-code/error-handling behavior is user-visible. I’ll reword that part.

Thanks, that clarification is fair.

I’m okay not blocking this PR on full usage/help standardization as the intent here is only to preserve prior behavior where it already existed. We can do the broader consistency cleanup in a follow-up PR.

But still:
- reword the PR description to remove the “no impact on the binary” claim
- fix the remaining unclassified error paths so they don’t fall through to generic exit code `20`

Caesarsage:
> > Thanks for the review. The usage/help preservation was only added on paths that already printed usage before this PR. So I wouldn’t call the remaining commands a consistency regression in this PR, since they did not previously have that behavior. I’m happy to raise a follow-up PR to standardize usage/help output across all commands. For this PR, my intent was narrower: migrate commands from Run to RunE and route returned errors through the new centralized handler.
> > I agree the PR description should not say “No impact on the binary” so strongly, since the exit-code/error-handling behavior is user-visible. I’ll reword that part.
> 
> Thanks, that clarification is fair.
> 
> I’m okay not blocking this PR on full usage/help standardization as the intent here is only to preserve prior behavior where it already existed. We can do the broader consistency cleanup in a follow-up PR.
> 
> But still:
> 
> * reword the PR description to remove the “no impact on the binary” claim
> * fix the remaining unclassified error paths so they don’t fall through to generic exit code `20`

done. Thank you

Caesarsage:
I checked the remaining paths and the real issue is not the KindUsage returns inside RunE; those already map to exit code 2.

The remaining unclassified paths are Cobra-generated errors that happen before RunE runs, so they fall through to generic 20:

- cmd/test.go: cobra.ExactArgs(3) -> microcks test
- cmd/import.go: cobra.MaximumNArgs(1) -> microcks import a b
- root/subcommand flag parsing -> unknown flags like microcks --bad
- cmd/cmd.go: MarkFlagsRequiredTogether(...) -> one Keycloak flag without the other

Vaishnav88sk:
> I checked the remaining paths and the real issue is not the KindUsage returns inside RunE; those already map to exit code 2.
> 
> The remaining unclassified paths are Cobra-generated errors that happen before RunE runs, so they fall through to generic 20:
> 
> * cmd/test.go: cobra.ExactArgs(3) -> microcks test
> * cmd/import.go: cobra.MaximumNArgs(1) -> microcks import a b
> * root/subcommand flag parsing -> unknown flags like microcks --bad
> * cmd/cmd.go: MarkFlagsRequiredTogether(...) -> one Keycloak flag without the other

Thanks!

So could you fix this in PR or separate?
Or should we keep it?

Caesarsage:
> > I checked the remaining paths and the real issue is not the KindUsage returns inside RunE; those already map to exit code 2.
> > The remaining unclassified paths are Cobra-generated errors that happen before RunE runs, so they fall through to generic 20:
> > 
> > * cmd/test.go: cobra.ExactArgs(3) -> microcks test
> > * cmd/import.go: cobra.MaximumNArgs(1) -> microcks import a b
> > * root/subcommand flag parsing -> unknown flags like microcks --bad
> > * cmd/cmd.go: MarkFlagsRequiredTogether(...) -> one Keycloak flag without the other
> 
> Thanks!
> 
> So could you fix this in PR or separate? Or should we keep it?

a separate PR would done

Vaishnav88sk:
> a separate PR would done

Okay then, please do it so it will be clean code.
Ready to merge @Harsh4902

Harsh4902:
@Caesarsage Have you already raised a seperate PR for  changes suggested by @Vaishnav88sk or you will work on that after merging?

Caesarsage:
> @Caesarsage Have you already raised a seperate PR for changes suggested by @Vaishnav88sk or you will work on that after merging?

After merging. With a couple of other PRs needed to used the CLI as vscode extension interface. Since the suggestions are nor blocker. Merging this now is a good idea

Harsh4902 (approved):
LGTM

PR #491, opened 2026-07-09
refactor(cli): remove the legacy exit shim and enforce it in CI

Caesarsage:
Part of the Microcks CLI v2 work (https://github.com/microcks/microcks-cli/issues/255). Stacked on PR https://github.com/microcks/microcks-cli/pull/490.

- Deletes the now-unused `CheckError`/`CheckConfigNil`/`Fatal` + `ErrorGeneric` shim from `pkg/errors`; converts the standalone `watcher/` binary to handle errors inline.
- Adds `go test ./...` to `build-verify.yml` (the pipeline previously only built) and a **guard step** that fails if `os.Exit`/`log.Fatal`/`panic` reappears outside `cmd/exit.go`.
- Documents the rule in `CONTRIBUTING.md`.

### Scope
Pure cleanup: nothing referenced the shim after PR https://github.com/microcks/microcks-cli/pull/490. 

The only process exits left in the tree are the two `main` entrypoints and `cmd.Handle`.

### DIFF
--- a/.github/workflows/policy-guards.yml
+++ b/.github/workflows/policy-guards.yml
@@ -1,0 +1,62 @@
+name: policy-guards
+
+on:
+  push:
+    paths:
+      - '**/*.go'
+      - '.github/workflows/policy-guards.yml'
+  pull_request:
+    paths:
+      - '**/*.go'
+      - '.github/workflows/policy-guards.yml'
+
+permissions:
+  contents: read
+
+jobs:
+  policy-guards:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd #v6.0.2
+
+      - name: Guard Go copyright headers
+        run: |
+          missing=""
+          while IFS= read -r file; do
+            if ! grep -q "Copyright The Microcks Authors" "$file"; then
+              missing+="${file}"$'\n'
+            fi
+          done < <(git ls-files '*.go')
+
+          if [[ -n "$missing" ]]; then
+            echo "::error::Go files missing Microcks copyright header:"
+            printf "%s" "$missing"
+            exit 1
+          fi
+
+      - name: Guard against silently ignored errors
+        run: |
+          # Avoid reintroducing review issues where malformed responses or local
+          # failures are discarded instead of being returned as classified errors.
+          if grep -rnE '(^|[^[:alnum:]_])_ =|, _ :=|fmt\.Println\(err\)' --include='*.go' cmd pkg; then
+            echo "::error::Potential silently ignored error found. Handle it explicitly or return errors.Wrap(kind, err)."
+            exit 1
+          fi
+
+      - name: Guard against stray process exits
+        run: |
+          # Library and command packages must return classified errors, never
+          # exit or panic. Only the explicit process entrypoints may exit.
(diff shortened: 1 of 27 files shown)


Caesarsage:
HOLD til https://github.com/microcks/microcks-cli/pull/490 and https://github.com/microcks/microcks-cli/pull/489 are review and merged

pkg/connectors/microcks_client.go
```diff
@@ -239,24 +239,29 @@ func (c *microcksClient) GetKeycloakURL() (string, error) {
...
-	realmName := configResp["realm"].(string)
+	// Return 'null' if Keycloak is disabled.
+	if enabled, _ := configResp["enabled"].(bool); !enabled {
+		return "null", nil
+	}
 
-	// Return a proper URL or 'null' if Keycloak is disables.
-	if enabled {
-		return authServerURL + "/realms/" + realmName + "/", nil
+	authServerURL, _ := configResp["auth-server-url"].(string)
```
Vaishnav88sk: Please check this as the same issue as previous PRs. This silently discards the malformed-response case again. Check all such cases to ensure code consistency.
Rebase

pkg/errors/error.go
Vaishnav88sk, line 1: Needs a rebase, so the previous copyright block will be there

Vaishnav88sk (dismissed):

pkg/connectors/microcks_client.go
Caesarsage: The empty-string check catches most malformed cases, but explicit ok is cleaner and I will implement that. Thanks

Caesarsage:
Hello @Vaishnav88sk and @Harsh4902 

I have attended to all the issues here and from the other PR (merged) here instead of raising a separate PR. And i also added a CI guard to check for missing policies, like missing copyright header, ignored-error and stray-exit

.github/workflows/policy-guards.yml
```diff
@@ -0,0 +1,62 @@
...
+          fi
+
+      - name: Guard against silently ignored errors
+        run: |
+          # Avoid reintroducing review issues where malformed responses or local
+          # failures are discarded instead of being returned as classified errors.
+          if grep -rnE '(^|[^[:alnum:]_])_ =|, _ :=|fmt\.Println\(err\)' --include='*.go' cmd pkg; then
+            echo "::error::Potential silently ignored error found. Handle it explicitly or return errors.Wrap(kind, err)."
+            exit 1
+          fi
```
Vaishnav88sk, line 45: The silently-ignored-error grep pattern `_ =|, _ :=|fmt\.Println\(err\)` is a nice guard, but it's a blunt regex; it'll false-positive on any deliberate `_ = x` (e.g. explicitly discarding a value that's genuinely fine to ignore, if that ever comes up) with no allowlist mechanism. Not blocking, just flag it as something that may need a `// nolint`-style escape hatch down the line if a legitimate case shows up.

Vaishnav88sk (commented):
@Harsh4902  Please review the workflows/ files.

Vaishnav88sk:
LGTM 👍🏻

Caesarsage:
@Harishrs2006 kindly take a look and approve

Caesarsage:
@Harsh4902

Harsh4902 (approved):
GTG. Thanks @Caesarsage

PR #494, opened 2026-07-25
fix: handle StringFromCharset against empty charset and add Unit test for rand package

### LINKED ISSUE #493: StringFromCharset panics on empty charset — pkg/util/rand has no test coverage
### Reason/Context

**Describe the bug**
`StringFromCharset(n, charset)` in `pkg/util/rand/rand.go` panics instead of returning an error when called with an empty charset and `n > 0`. This package also currently has zero unit test coverage, which is how this went unnoticed.

**To Reproduce**
```go
_, err := rand.StringFromCharset(5, "")
```

This panics with:
`panic: crypto/rand: argument to Int is <= 0`

### Description

Root cause: `maxIdx := big.NewInt(int64(len(charset)))` evaluates to `0` when `charset` is empty, and that value is passed directly into `crypto/rand.Int(rand.Reader, maxIdx)`, which panics whenever `max <= 0`.

**Expected behavior**
`StringFromCharset` should return `("", error)` for this input, consistent with how the function already handles other internal failures — errors from `crypto/rand.Int` elsewhere in the loop are wrapped and returned rather than left to panic.

**Additional context**
`pkg/util/rand` has no test file at all today. `String()` and `StringFromCharset()` back the OAuth2 state / PKCE value generation used in `cmd/login.go`, so both the missing coverage and this panic path are worth closing together — the fix without tests would just leave the same class of bug free to reappear.

I'd like to work on this: fixing the panic and adding full unit test coverage for the package in the same change. 

### Implementation ideas

_No response_

richochetclementine1315:
## What
Fixes a panic in `StringFromCharset(n, charset)` — it panicked with `crypto/rand: argument to Int is <= 0` when called with an empty charset and `n > 0`, instead of returning an error the way its other failure paths already do. Also adds the first unit test coverage for `pkg/util/rand`, which previously had none.

## Why
Closes #493 

## Changes
- `pkg/util/rand/rand.go`: guard against an empty charset before calling `crypto/rand.Int`, returning a descriptive error instead of panicking.
- `pkg/util/rand/rand_test.go`: table-driven tests for `String` and `StringFromCharset`, covering:
  - output length matches requested length across zero/single/short/long inputs
  - output characters all belong to the supplied/default charset
  - two calls with identical parameters produce different output
  - single-character charset behavior
  - empty-charset behavior — the case that previously panicked and now correctly returns an error

## Testing
- `go test ./pkg/util/rand/... -v` — all cases pass, including the previously-panicking empty-charset case.
- `go test ./...` — full suite passes.
Screenshot 2026-07-25 222139

## Notes
No behavior change for any existing valid input — `String()` and existing callers of `StringFromCharset()` with a non-empty charset are unaffected. This only changes behavior for the previously-panicking empty-charset edge case, and adds coverage that didn't exist before.

### DIFF
--- a/pkg/util/rand/rand.go
+++ b/pkg/util/rand/rand.go
@@ -15,6 +15,9 @@ func String(n int) (string, error) {
 
 // StringFromCharset generates, from a given charset, a cryptographically-secure pseudo-random string of a given length.
 func StringFromCharset(n int, charset string) (string, error) {
+	if n > 0 && len(charset) == 0 {
+		return "", fmt.Errorf("charset must not be empty when n > 0")
+	}
 	b := make([]byte, n)
 	maxIdx := big.NewInt(int64(len(charset)))
 	for i := 0; i < n; i++ {
@@ -22,7 +25,6 @@ func StringFromCharset(n int, charset string) (string, error) {
 		if err != nil {
 			return "", fmt.Errorf("failed to generate random string: %w", err)
 		}
-		// randIdx is necessarily safe to convert to int, because the max came from an int.
 		randIdxInt := int(randIdx.Int64())
 		b[i] = charset[randIdxInt]
 	}
--- a/pkg/util/rand/rand.go
+++ b/pkg/util/rand/rand.go
@@ -15,6 +15,9 @@ func String(n int) (string, error) {
 
 // StringFromCharset generates, from a given charset, a cryptographically-secure pseudo-random string of a given length.
 func StringFromCharset(n int, charset string) (string, error) {
+	if n<0{
+		return "", fmt.Errorf("n must be greater than 0")
+	}
 	if n > 0 && len(charset) == 0 {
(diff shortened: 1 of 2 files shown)


richochetclementine1315:
@Vaishnav88sk  @lbroudoux @yada @Harsh4902 requesting to review and suggest changes if needed

Vaishnav88sk:
LGTM @Harsh4902

Caesarsage:
+1 — verified the guards cover both reported panics: the empty-charset case, and via the n<0 check the String(-1) case from #359 (String delegates to StringFromCharset). One nit: gofmt the new guard lines (if n<0{). When this merges, #359, #493 and #360 can all close against it — #360 predates this with the same idea, so a credit mention there would be a kind touch.

Harsh4902 (approved):
GTG

richochetclementine1315:
Thankyou Everyone!!...Looking forward to making more meaningful and quality contributions at MicrocksCLI...

PR #495, opened 2026-07-26
feat: add microcks validate command for pre-flight config verification

### LINKED ISSUE #470: feat: Add microcks validate command for pre-flight config verification
### Current Behavior
Users discover configuration problems only when running actual commands. Common issues include expired tokens, unreachable URLs, missing permissions, or incompatible TLS settings. There is no way to validate the entire setup before attempting imports or tests.

### Why we need this?
- Config errors surface mid-execution, wasting time and CI resources
- New users struggle with silent misconfigurations
- No pre-flight check exists for CI/CD pipelines

### How will this help?
- `microcks validate` catches issues before imports/tests run
- Structured pass/fail output enables CI gating
- Reduces onboarding friction for new users
- Prevents failed CI runs due to expired tokens or wrong contexts

### Proposed Solution
Add `microcks validate` command that performs:
- Config file existence and permissions check
- Current context validity verification
- Server reachability (HTTP 200 response)
- Keycloak token validity and expiration check
- Service account permissions (manager/admin roles)
- Async Minion connectivity status

Exit codes: 0 (all passed), 1 (one or more failed)

Example output:
```bash
$ microcks validate
✓ Config file exists at /home/user/.config/microcks/config
✓ Current context 'staging' is valid
✓ Microcks server at http://localhost:8585/api is reachable (HTTP 200)
✓ Keycloak authentication is enabled and token is valid
✓ Service Account has 'manager' role for imports
✓ Async Minion is connected

All checks passed — ready to import and test

### PR DESCRIPTION
**Pull request by A-d-i-t-y**
Fixes #470

## Summary
Adds a `microcks validate` command that runs pre-flight checks before 
running imports or tests, so users catch configuration problems early 
instead of mid-execution.

## What it checks
- Config file is present and parses correctly
- A context is configured and resolvable
- Auth token is valid (refreshed automatically if needed)
- The target Microcks server is reachable

Exits with code `0` if all checks pass, `1` if any check fails — 
suitable for CI/CD pre-flight gating.

### DIFF
--- a/cmd/validate.go
+++ b/cmd/validate.go
@@ -25,79 +25,120 @@ import (
 )
 
 // NewValidateCommand builds the "validate" command which runs pre-flight
-// checks on the local configuration and connectivity to a Microcks server,
-// without performing any import or test action.
+// checks on configuration and connectivity to a Microcks server, without
+// performing any import or test action.
 func NewValidateCommand(globalClientOpts *connectors.ClientOptions) *cobra.Command {
 
 	var validateCmd = &cobra.Command{
 
 		Use:   "validate",
 		Short: "Validate current Microcks CLI configuration and connectivity",
-		Long: `Runs pre-flight checks on the active Microcks context:
-   - config file is present and parses correctly
-   - a context is configured and resolvable
-   - the auth token is valid (refreshed if needed)
+		Long: `Runs pre-flight checks before running imports/tests:
+   - config/credentials are resolvable, either via a local context or via
+	 --microcksURL together with --keycloakClientId/--keycloakClientSecret
+	 for CI/CD (service account) usage
+   - the auth token is valid (refreshed / retrieved as needed)
    - the target Microcks server is reachable
  
- Exits with code 0 if all checks pass, or 1 if any check fails.`,
+ Exits with code 0 if all checks pass, or 1 if any check fails.
+ 
+ Note: this command validates connectivity and authentication only. It
+ does not check service-account authorization scopes or Async API Minion
+ connectivity, which are tracked as a follow-up.`,
 		Example: `microcks validate
- microcks validate --microcks-context staging`,
+ microcks validate --microcks-context staging
+ microcks validate --microcksURL http://microcks.example.com/api --keycloakClientId my-sa --keycloakClientSecret my-secret`,
 
 		Run: func(cmd *cobra.Command, args []string) {
-			ok := true
-
-			// 1. Config file presence and parsing.
-			configFile := globalClientOpts.ConfigPath
-			localConfig, err := config.ReadLocalConfig(configFile)
-			if err != nil {
-				fmt.Printf("✗ Config file: invalid — %v\n", err)
-				os.Exit(1)
-			}
-			fmt.Println("✓ Config file: parsed successfully")
-
-			if localConfig == nil || localConfig.IsEmpty() {
-				fmt.Println("✗ No contexts configured. Run 'microcks login <server>' first.")
-				os.Exit(1)
-			}
-
-			// 2. Context resolution.
-			ctx, err := localConfig.ResolveContext(globalClientOpts.Context)
-			if err != nil {
-				fmt.Printf("✗ Context resolution: %v\n", err)
-				os.Exit(1)
-			}
-			fmt.Printf("✓ Context resolved: %q (server: %s)\n", ctx.Name, ctx.Server.Server)
-
-			// 3. Build a client for this context. NewClient() attempts a token
-			// refresh internally, so a failure here reveals an expired or
-			// invalid credential.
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by A-d-i-t-y** [comment]
Hi @Harsh4902 @Vaishnav88sk

I've opened a PR for this issue: #495

It adds a 'microcks validate' command that runs pre-flight checks 
(config validity, context resolution, auth token, server reachability) 
before running imports/tests, as described in the issue. Included unit 
tests and manually verified the build and CLI output.

I'm applying for the LFX Mentorship (Term 3) and would really appreciate 
your review whenever you get a chance. Happy to make any changes needed!

**Comment by Vaishnav88sk** [comment]
> Hi @Harsh4902 @Vaishnav88sk
> 
> I've opened a PR for this issue: #495
> 
> It adds a 'microcks validate' command that runs pre-flight checks (config validity, context resolution, auth token, server reachability) before running imports/tests, as described in the issue. Included unit tests and manually verified the build and CLI output.
> 
> I'm applying for the LFX Mentorship (Term 3) and would really appreciate your review whenever you get a chance. Happy to make any changes needed!

Thanks for the PR.
We are not sure about Microcks' participation in Term 3.

File: cmd/validate.go [hunk 1]
```diff
@@ -0,0 +1,103 @@
...
+		Run: func(cmd *cobra.Command, args []string) {
+			ok := true
+
+			// 1. Config file presence and parsing.
+			configFile := globalClientOpts.ConfigPath
+			localConfig, err := config.ReadLocalConfig(configFile)
+			if err != nil {
+				fmt.Printf("✗ Config file: invalid — %v\n", err)
+				os.Exit(1)
+			}
```
Vaishnav88sk: `ReadLocalConfig()` can return `(nil, nil)` for a missing file, but we print `✓ Config file: parsed successfully` before checking whether any config/context actually exists. Please restructure this so the success message is only printed when there is a real config to validate.

cmd/validate.go
```diff
@@ -0,0 +1,103 @@
...
+			mc, err := connectors.NewClient(*globalClientOpts)
+			if err != nil {
+				fmt.Printf("✗ Authentication: %v\n", err)
+				ok = false
+			} else {
+				fmt.Println("✓ Authentication: token is valid")
+			}
+
+			// 4. Server reachability check (also confirms Keycloak config
+			// endpoint responds, covering both plain connectivity and
```
Vaishnav88sk: This currently only validates the local config/context path.

Please also consider the direct connection mode that other commands support via `--microcksURL` together with `--keycloakClientId` / `--keycloakClientSecret`. For CI/CD pre-flight checks, that mode is at least as important as local config validation.

cmd/validate.go
```diff
@@ -0,0 +1,103 @@
...
+
+		Use:   "validate",
+		Short: "Validate current Microcks CLI configuration and connectivity",
+		Long: `Runs pre-flight checks on the active Microcks context:
+   - config file is present and parses correctly
+   - a context is configured and resolvable
+   - the auth token is valid (refreshed if needed)
+   - the target Microcks server is reachable
+ 
+ Exits with code 0 if all checks pass, or 1 if any check fails.`,
```
Vaishnav88sk: The command contract in #470 is broader than what is currently implemented.

Right now this checks config/context/auth/server reachability, but it does not implement the proposed service-account permission check or Async Minion connectivity check. Please either extend the implementation to cover those, or narrow the issue/PR scope explicitly.

cmd/validate_test.go
Vaishnav88sk, line 1: These tests do not actually exercise `NewValidateCommand()` or the command output/exit behavior.

They mostly re-test `config.ReadLocalConfig()` and `ResolveContext()`, which are existing helpers. Please add command-level tests for the new behavior itself: success path, missing config/context, and a failing auth/reachability path if possible.

Vaishnav88sk: Add a copyright block

Vaishnav88sk (changes requested):
Pass the DCO

A-d-i-t-y:
Thanks for the detailed review, @Vaishnav88sk! Pushed a new commit (df3dfde) addressing all the points:

- Fixed the message-ordering bug (config check now happens before printing success)
- Added CI/CD direct-connection mode (--microcksURL + keycloak client creds), following the same pattern as cmd/test.go
- Rewrote the tests to exercise runValidate() directly (success, missing-config, invalid-config, unreachable-server, and direct-mode-fallback paths)
- Added copyright header to the test file
- Narrowed scope explicitly in --help text regarding service-account/Async Minion checks (replied inline with details)

Ready for another look whenever you get a chance!

Vaishnav88sk:
Thanks for the PR!
But I am not sure about the use case for this. Where is it needed?

Could you show some logs or the output of the command?

Issue #503, opened 2026-08-17
fix(security): credential redaction in 0f8b000 misses JSON bodies access_token, refresh_token, clientSecret, and password print verbatim with --verbose


gyanranjanpanda:
## Problem

 i added credential redaction to verbose HTTP dumps. It works for `Authorization` headers but misses both places where secrets actually appear.

**Leak 1  Keycloak token response** (`keycloak_client.go:96`)

The response body is JSON. The regex expects `name=value` (form encoding), so `"access_token":"eyJ..."` prints verbatim.

**Leak 2  create-test request body** (`microcks_client.go:371`)

`OAuth2ClientContext` is marshalled into the POST body. Fields `clientSecret`, `password`, and `refreshToken` aren't in the pattern at all.

The header redaction fires correctly, which is exactly what makes this easy to miss:
```
Authorization: [REDACTED]       ← works
{"oAuth2Context":{"clientSecret":"SECRET","password":"SECRET",...}}   ← leaks
```

## Root Cause

```go
// only matches name=value — never fires on JSON bodies
var sensitiveParamPattern = regexp.MustCompile(
    `(?i)(access_token|refresh_token|id_token|code)=([^&\s]+)`,
)
```

Apply it alongside the existing patterns in `redactSensitiveContent`.

## Reproduce

Drop into `pkg/config/redact_probe_test.go` and run `go test ./pkg/config/ -run TestRedactProbe -v` — 6 FAILs, all confirmed against the real function.

gyanranjanpanda:
@Harsh4902 @Caesarsage could verify this issue

Harsh4902:
Thanks @gyanranjanpanda, I think this is correct. As we are storing the credentials into the `config` file, we are exposing the AuthClientContext, and this can lead us to leaking credentials. Though I'll do the thorough check.

Vaishnav88sk:
Thanks for reporting this @gyanranjanpanda, you are completely right. The current `sensitiveParamPattern` regex only covers form-urlencoded payloads and completely misses JSON bodies, which is a genuine security leak when using `--verbose`.

@Harsh4902 - I think that this issue is specifically about the HTTP request/response dumps logging secrets to stdout during `--verbose`, rather than the config file storage.

**Proposed Solution:** We should add an additional regex pattern to `redactSensitiveContent` specifically for JSON payloads, such as: `(?i)" (access_token|refresh_token|id_token|code|clientSecret|password)"\s*:\s*"([^"]+)"`

[...]

PR #504, opened 2026-08-17
fix(security): redact credentials in JSON and form bodies under --verbose

### LINKED ISSUE #449: Security: SSO login leaks OAuth access and refresh tokens to CLI logs
## Summary

After a successful SSO login, `oauth2login()` in `cmd/login.go` unconditionally logs both the OAuth access token and refresh token to stderr:

```go
// cmd/login.go, lines 296-297
log.Printf("Token: %s\n", tokenString)
log.Printf("Refresh Token: %s\n", refreshToken)
```

This is a **critical credential leak**. The refresh token is long-lived and can be used to mint new access tokens. These logs can end up in:
- Terminal scrollback / shell history
- CI/CD pipeline logs
- Support bundles or shared debug output
- Shell capture tools (e.g. `script`, `tee`)

A user only needs to run:
```bash
microcks login <server> --sso
```
and the CLI exposes the full session credentials after authentication.

## Additional Leaks Found

| Leak | Location | Severity | Content |
|------|----------|----------|---------|
| Access token | `cmd/login.go:296` | **Critical** | Full JWT access token |
| Refresh token | `cmd/login.go:297` | **Critical** | Long-lived refresh token |
| Callback URL with auth code | `cmd/login.go:222` | **High** | OAuth authorization code via `log.Printf("Callback: %s\n", r.URL)` |

The callback URL log at line 222 also leaks the OAuth authorization code, which (though short-lived) can be exchanged for tokens if intercepted before the CLI completes the exchange.

## Why This Is Critical

The project already stores tokens securely in a protected config file (`~/.microcks/microcks.yaml`), but **these log lines bypass that protection entirely** by printing secrets directly to the terminal. This logging is:
- **Not gated** behind `--verbose` or any debug flag
- **Always active** on every SSO login
- Exposing **both** immediate access (access token) and long-term session hijack capability (refresh token)

## Suggested Fix

1. **Remove** the two `log.Printf` calls at lines 296-297 that print the access and refresh tokens
2. **Remove or redact** the callback URL log at line 222, or strip query parameters (which contain the OAuth `code`)
3. **Audit** any verbose/debug HTTP dump paths to ensure `Authorization`, `access_token`, `refresh_token`, and `id_token` values are redacted
4. Consider gating all OAuth-related debug logging behind an explicit `--verbose` flag

## Reproduction

```bash
microcks login http://localhost:8080 --sso
# After browser-based authentication completes, observe stderr output containing:
# Token: eyJhbGciOi...
# Refresh Token: eyJhbGciOi...
```

## Environment

- microcks-cli: current main branch
- File: `cmd/login.go`
- Lines: 222, 296-297

gyanranjanpanda:
## Problem

Commit `0f8b000` (closes #449) added `redactSensitiveContent` to verbose HTTP dumps. It works for `Authorization` headers but the `sensitiveParamPattern` regex only matches `name=value` (form encoding). Both dump sites that carry secrets produce **JSON bodies**, so the pattern never fires.

**Leak 1** — `keycloak_client.go:96`: Keycloak token response body is JSON. `access_token`, `refresh_token`, and `id_token` print verbatim.

**Leak 2** — `microcks_client.go:371`: `OAuth2ClientContext` is marshalled into the POST `/api/tests` body. `clientSecret`, `password`, and `refreshToken` are not in the pattern at all.

The `Authorization: [REDACTED]` header line makes the output look safe, masking the residual exposure.

Tracked in #503.

## Fix

Replace the single catch-all regex with a structured approach:

- **JSON bodies** — parse with `encoding/json`, redact sensitive keys by name
- **Form bodies** — parse with `url.ParseQuery`, redact by key name
- **Unknown/chunked** — text-pattern fallback

Key matching normalises `snake_case`, `camelCase`, and `kebab-case` so every spelling variant is covered (`access_token`, `accessToken`, `Access-Token` all match).

Non-sensitive metadata (`token_type`, `expires_in`, `serviceId`) is preserved so `--verbose` stays useful.

## Tests

`pkg/config/config_test.go` — 13 tests (all new), covering:
- JSON token response (regression for Leak 1)
- JSON test request with `oAuth2Context` (regression for Leak 2)
- Form-encoded token exchange
- Nested/array JSON
- Chunked transfer fallback
- Header-only dumps
- Non-sensitive body preservation
- CRLF round-trip
- Key normalisation variants
- End-to-end through `DumpRequestIfRequired` / `DumpResponseIfRequired`

```
ok  github.com/microcks/microcks-cli/pkg/config  0.657s
```


pkg/config/config_test.go
```diff
@@ -456,3 +456,202 @@ func TestWatchConfig(t *testing.T) {
...
+			}
+			for _, unwanted := range tt.mustNotHave {
+				assert.NotContains(t, got, unwanted)
+			}
+		})
+	}
+}
+
+func TestRedactSensitiveContentPreservesCRLF(t *testing.T) {
+	dump := "GET / HTTP/1.1\r\nAuthorization: Bearer eyJLEAKEDACCESS\r\nAccept: */*\r\n\r\n"
```
Caesarsage, line 590: One gap here: authToken / auth-token values leak through — authtoken isn't in sensitiveValueKeys, and auth-token is the exact key the CLI's own config uses:

dump := "HTTP/1.1 200 OK\r\nContent-Type: application/json\r\n\r\n" +
    `{"authToken":"eyJPROBELEAK"}`
redactSensitiveContent(dump) // value survives
Adding "authtoken": {} to the set covers both spellings via the normalizer.

Minor, non-blocking: json.Marshal re-encoding reorders keys and HTML-escapes, so dumped bodies aren't byte-faithful (and won't match Content-Length) — worth a one-line code comment.

Caesarsage (commented):
This is good.

I left a comment. Also, could you reduce the comments , it is making the code unnecessarily large and noisy. comments is only useful where it is really needed like on the complex regex and should be simple

gyanranjanpanda:
This looks good to go from my end and seems ready to merge. @Caesarsage @Harsh4902, could you share your views on this?

PR #510, opened 2026-08-27
  Fix: import command to honor `--microcksURL` without login

### LINKED ISSUE #511: import command requires login even when --microcksURL is provided
### Describe the bug

  `microcks import` does not seem to handle `--microcksURL` the same way as `import-dir`.

  I started a local Microcks instance using `microcks start`, which runs the `latest-native` image without Keycloak enabled. In this setup, importing through `--microcksURL
  http://localhost:8585` should work without a saved login/config.

  `import-dir` works correctly with an empty config path and imports the artifact successfully. But `import` with the same server URL and same empty config path fails with:

  ```
  please login to perform this operation
```
  This is confusing because the command already has the target server from `--microcksURL`.

### Expected behavior

microcks import `--microcksURL`... should try to use the provided server directly, like `--import-dir` does. If the server requires authentication, then it
  should fail with an auth/API error from the server.

  The issue seems to be that `cmd/import.go` has its own client setup logic, while import-dir uses the shared `newCommandClient(globalClientOpts)` helper.

### Repro

Image

  microcks-cli master branch

_No response_

### Additional information

_No response_

### PR DESCRIPTION
**Pull request by Syedowais312**
## Summary

  This updates `microcks import` to use the shared command client setup already used by `import-dir`, `service`, and `test`.

  Previously, `microcks import --microcksURL ...` still required local config/login unless Keycloak client credentials were also provided. This was inconsistent with  `import-dir` and broke local `latest-native` usage where Keycloak is disabled.

  ## Changes

  - Reuse `newCommandClient(globalClientOpts)` in the `import` command.
  - Preserve `--watch` context selection by falling back to the current config context, then the provided server URL.
  - Add a regression test verifying `import --microcksURL --config <empty>` reaches the artifact upload endpoint.

  ## Verification

  ```bash
  go test ./cmd
  go test ./...

```

fix: #511

### DIFF
--- a/cmd/import.go
+++ b/cmd/import.go
@@ -52,71 +52,14 @@ func NewImportCommand(globalClientOpts *connectors.ClientOptions) *cobra.Command
 
 			specificationFiles := args[0]
 
-			// Initialize config from command options.
-			config.InsecureTLS = globalClientOpts.InsecureTLS
-			config.CaCertPaths = globalClientOpts.CaCertPaths
-			config.Verbose = globalClientOpts.Verbose
-
-			// Read local config file in case we need some context info.
-			localConfig, err := config.ReadLocalConfig(globalClientOpts.ConfigPath)
+			mc, serverAddr, err := newCommandClient(globalClientOpts)
 			if err != nil {
-				return errors.Wrap(errors.KindEnvironment, err)
+				return err
 			}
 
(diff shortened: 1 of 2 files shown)


Syedowais312:
cc: @Vaishnav88sk @Harsh4902 @lbroudoux

PR #527, opened 2026-09-17
refactor: remove redundant HasPrefix check before TrimPrefix

### LINKED ISSUE #526: refactor: remove redundant `strings.HasPrefix` check before `strings.TrimPrefix`
### Describe the bug

In `cmd/import.go`, there is an `if strings.HasPrefix(...) { ... = strings.TrimPrefix(...) }` check. This is redundant boilerplate. `strings.TrimPrefix` is safe to call unconditionally because it simply returns the original string if the prefix isn't found. 

### Expected behavior

The `strings.TrimPrefix` function should be called without the surrounding `if` statement to reduce cyclomatic complexity.

### Actual behavior

The codebase contains a redundant `if` statement that adds unnecessary lines of code.

### How to Reproduce?

Reproducer:
1. Look at `cmd/import.go`, specifically where path prefixes are trimmed.
2. Observe the `if strings.HasPrefix(...)` check guarding the `strings.TrimPrefix` call.

`main` (Latest)

CLI

### Additional information

This is a simple `good first issue` to clean up some Go syntax. Removing the `if` block and unconditionally calling `strings.TrimPrefix` fulfills the same logic cleanly and passes static checkers.

ish-g09:
### Description
1. Removed the redundant if 'strings.HasPrefix(f, "./")' check in 'cmd/import.go'
2. 'strings.TrimPrefix' already returns the string unchanged if the prefix is not found, so the 'if' block was unnecessary
3. This reduces cyclomatic complexity with no change in behavior
4. All existing tests pass

### Related issue(s)
Fixes #526

### AI Disclosure
The code change was written and verified by me, I fully understand it.


ish-g09:
Closing this PR as it was accidentally opened from the master branch. The issue https://github.com/microcks/microcks-cli/issues/526 fix will be submitted in a clean dedicated branch PR shortly.

PR #528, opened 2026-09-17
Fix/replace deprecated functions

### LINKED ISSUE #525: chore: replace deprecated function usages (`filepath.HasPrefix` and `client.IsErrNotFound`)
### Describe the bug

The codebase contains a few function calls that have been deprecated by their respective upstream libraries. Continuing to use them creates technical debt and makes future library upgrades difficult.
- `pkg/connectors/container_client.go:232`: Uses `client.IsErrNotFound(err)` from the Docker SDK, which is deprecated.
- `cmd/importDir_test.go:71`: Uses `filepath.HasPrefix(path, root)`, which has been deprecated since Go 1.0 because it does not respect path boundaries.

### Expected behavior

The codebase should use modern, supported equivalents for these operations.

### Actual behavior

Deprecated functions are used, triggering linter warnings (`SA1019`).

### How to Reproduce?

Reproducer:
1. Run `staticcheck ./...` locally.
2. Observe the deprecation warnings (`SA1019`) for both functions.

`main` (Latest)

CLI

### Additional information

This is a `good first issue`. 
- `client.IsErrNotFound(err)` should be replaced with `cerrdefs.IsNotFound(err)`.
- `filepath.HasPrefix` should be replaced with `strings.HasPrefix(path, root+string(filepath.Separator))`.

ish-g09:
### Description

- .Replaced deprecated `client.IsErrNotFound(err)` with `cerrdefs.IsNotFound(err)` in `pkg/connectors/container_client.go`
- Replaced deprecated `filepath.HasPrefix(path, root)` with `path == root || strings.HasPrefix(path, root+string(filepath.Separator))` in `cmd/importDir_test.go` to properly handle path boundaries and root directory traversal
- Both were triggering `SA1019` linter warnings from staticcheck
- Added `cerrdefs "github.com/docker/docker/errdefs"` import

### Related issue(s)
Fixes #525

### AI Disclosure
The code changes were written and verified by me. I fully understand them.

### Testing
- `go build ./...` : passes
- `go test ./...`  : all packages pass


cmd/import.go
```diff
@@ -173,9 +173,7 @@ func NewImportCommand(globalClientOpts *connectors.ClientOptions) *cobra.Command
 					}
 
 					// Normalize file path to match the watcher fsnotify events format.
-					if strings.HasPrefix(f, "./") {
-						f = strings.TrimPrefix(f, "./")
-					}
+					f = strings.TrimPrefix(f, "./")
```
Syedowais312: The changes in this file belong to #526 and duplicate of #527

ish-g09: You're right, sorry about this. I had accidentally committed the issue #526 fix directly to master before creating this branch, so it got carried over here unintentionally.

cmd/importDir_test.go
```diff
@@ -68,7 +69,7 @@ func (m *MockFileSystem) Walk(root string, walkFn filepath.WalkFunc) error {
 	}
 
 	for path, isDir := range m.Files {
-		if filepath.HasPrefix(path, root) {
+		if strings.HasPrefix(path, root+string(filepath.Separator)) {
```
Vaishnav88sk: This introduces a logical failure during traversal. When `filepath.Walk` visits the root directory itself, `path` is exactly equal to `root`. Evaluating `strings.HasPrefix("root", "root/")` will return `false`, causing the root directory to be skipped entirely. Suggestion:
```suggestion
        if path == root || strings.HasPrefix(path, root+string(filepath.Separator)) {
```

Vaishnav88sk (dismissed):

cmd/importDir_test.go
ish-g09: Thanks for catching that, @Vaishnav88sk! That's a great point , the root directory would have been skipped otherwise. I've updated the check to include path == root as suggested.

Vaishnav88sk:
Now it looks good. can safely merge it. 👍🏻

Harsh4902 (approved):
LGTM. Thanks @ish-g09 for this PR.

PR #529, opened 2026-09-17
fix: parseImportURLArg correctly handles boolean-like secrets

### LINKED ISSUE #524: bug: `import-url` drops secret and breaks URL if secret resembles a boolean
### Describe the bug

When using the `import-url` command with a URL that includes a `primary` flag and a `secret`, the internal `parseImportURLArg` logic splits the string by `:` and searches backward for the boolean flag. If the provided secret is a string that parses as a boolean (e.g. `true`, `false`, `1`, `0`), the parser incorrectly consumes the secret as the `primary` flag. 

### Expected behavior

The command should correctly parse the URL, the boolean `primary` flag, and the `secret` string, even if the secret text happens to resemble a boolean value.

### Actual behavior

The secret is dropped entirely, the boolean `primary` flag is ignored, and the parsed URL incorrectly retains the boolean string at the end, causing the import to fail.

### How to Reproduce?

Reproducer:
1. Run a dry run or import using a boolean-like secret: 
   `microcks import-url "http://example.com/api:true:false"`
2. Observe that the logic parses the URL as `http://example.com/api:true`, the primary flag as `false`, and the secret as empty.

`main` (Latest)

CLI

### Additional information

This is a great `good first issue` to refactor the `parseImportURLArg` function in `cmd/importURL.go` to safely separate the URL, flag, and secret without relying on backward boolean parsing.

ish-g09:
### Description
- The `parseImportURLArg` function in `cmd/importURL.go` had a bug in how it parsed the `<URL>:<primary>:<secret>` argument format.

What actually was the bug: 
- The old logic scanned the URL parts backwards and treated the first boolean-parseable value it found as the `primary` flag. This broke when the `secret` itself looked like a boolean (e.g. `"true"`, `"false"`, `"1"`, `"0"`), or when the `secret` contained colons (e.g. `"my:secret:token"`).

_For example: `http://example.com/api:true:false`_
The old parser would find `"false"` first (scanning from the right), set `primary=false`, leave `secret` empty, and return the URL as `http://example.com/api:true` : all wrong.

The fix:
Instead of scanning backward, the new logic scans forward from index 2 to find the first boolean value representing the `primary` flag:
- The first segment that parses as boolean is set as the `primary` flag
- Everything before this index is reconstructed as the URL
- Everything after this index is joined with `:` to reconstruct the `secret`

This correctly parses all combinations of URL, primary flag, and secret, including secrets that resemble boolean values and secrets containing colons.
New test cases were added to `importURL_test.go` to cover these scenarios.

### Related issue(s)
Fixes #524

### AI Disclosure
I used AI assistance to understand the codebase. The code fix and test cases were written and verified by me. 
I fully understand the logic and can explain it to reviewers.

### Testing
- Added tests covering boolean-like secrets and secrets with colons
- Fixed `parseImportURLArg` tests now pass
- `go build ./...` : passes


cmd/importURL.go
```diff
@@ -73,16 +73,19 @@ func parseImportURLArg(f string) (string, bool, string) {
...
-					secret = strings.Join(parts[i+1:], ":")
-				}
-				f = strings.Join(parts[:i], ":")
-				break
+				secret = parts[n-1]
+				f = strings.Join(parts[:n-2], ":")
+			} else if val, err := strconv.ParseBool(parts[n-1]); err == nil {
+				// Case 2: URL:boolean
+				mainArtifact = val
+				f = strings.Join(parts[:n-1], ":")
```
Vaishnav88sk: This implementation strictly assumes the boolean flag is at either the `n-1` or `n-2` index. If a user passes a secret containing colons (e.g., `http://example.com/api:true:my:secret:token`), `strings.Split` breaks the string into `[http, //..., true, my, secret, token]`. The boolean (`true`) sits at `n-4`. Since this code only checks `parts[n-2]` (`secret`) and `parts[n-1]` (`token`), `strconv.ParseBool` fails on both, causing the parser to completely drop the flag and secret.
**Suggestion**: You must iterate backward to locate the boolean flag index `i`. Once located, reconstruct the URL from `parts[:i]` and the secret using `strings.Join(parts[i+1:], ":")`.

Vaishnav88sk (changes requested):

cmd/importURL.go
Syedowais312: Valid point, @Vaishnav88sk. I don't think we need to iterate backwards here, though.

We could instead scan forward from the first possible boolean position:

so basically instead of this
`for i := n - 1; i >= 2; i-- {`

we can do this

`for i := 2; i < n; i++ {`

and stop at the first value that successfully parses with `strconv.ParseBool`.

This also keeps the URL intact when it contains additional : characters, since everything before the boolean remains part of the URL. We can then reconstruct the URL with:

`f = strings.Join(parts[:i], ":")`

and treat `parts[i+1:] ` as the secret.

This is also consistent with the documented URL:boolean[:secret] format.
https://github.com/microcks/microcks-cli/blob/master/documentation/cmd/importURL.md#usage

ish-g09: Thank you both for the great feedback!

@Vaishnav88sk  Thanks for pointing out the colon separated secret edge case (my:secret:token), that was a blind spot in my previous implementation. 
@Syedowais312 The forward-scan approach from index 2 is great. It handles secrets with colons while also preventing boolean like secrets (from #524) from being mistakenly parsed as the primary flag. 

I will implement the forward-scan approach shortly along with comprehensive test cases covering both boolean like secrets and secrets containing colons, and update the PR.

Harsh4902: Resloving this as @ish-g09 has solved this one.

cmd/importURL_test.go
Vaishnav88sk, line 1: There are some unnecessary formatting chnages. please revert the formatting changes.

Vaishnav88sk (dismissed):

cmd/importURL_test.go
ish-g09: Did the needful. Thanks @Vaishnav88sk

Vaishnav88sk:
LGTM

Harsh4902 (dismissed):
LGTM. Thanks @ish-g09 for the PR.

Harsh4902 (approved):
GTG

Harsh4902 (dismissed):

Harsh4902:
@ish-g09 can you please have a look at the merge conflict?

ish-g09:
@Harsh4902 kindly check. Conflicts resolved

cmd/importURL.go
```diff
@@ -73,7 +73,7 @@ func parseImportURLArg(f string) (string, bool, string) {
 		parts := strings.Split(f, ":")
 		n := len(parts)
 
-		for i := n - 1; i >= 2; i-- {
+		for i := 2; i < n; i++ {
```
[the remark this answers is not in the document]
Vaishnav88sk: @ish-g09 ptal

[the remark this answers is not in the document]
ish-g09: Hey! I looked into Copilot’s comment and tested out its exact example (http://true:8080/api:false:secret).

Copilot’s trace was actually mistaken, when splitting by ':', parts[1] is '//true' and parts[2] is '8080/api'. Neither of those parses as a boolean, so the loop moves right past them to parts[3] (false), which correctly gets picked up as the primary flag while keeping the URL as http://true:8080/api.

I ran it locally through our test suite and it passes without any issues.

All CI checks are green (4/4). Happy to add this as an extra test case to importURL_test.go if you'd like, or we can keep it as is!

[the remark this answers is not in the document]
Harsh4902: Yes, this is a good finding. Please have a look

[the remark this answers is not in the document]
ish-g09: Hey @Harsh4902 , just added a test case in importURL_test.go covering that exact scenario (http://true:8080/api:false:secret). Tested it locally and pushed, it passes cleanly! Let me know what you think.

Harsh4902 (approved):
Good to go. Thanks @ish-g09

PR #532, opened 2026-09-17
refactor: remove redundant HasPrefix check before TrimPrefix

### LINKED ISSUE #526: refactor: remove redundant `strings.HasPrefix` check before `strings.TrimPrefix`

ish-g09:
###Description
- Removed the redundant if 'strings.HasPrefix(f, "./")' check in 'cmd/import.go'
- Replaced `strings.TrimPrefix` with `filepath.Clean(f)` to provide robust path normalisation (handling redundant prefixes like `././` and platform-specific path separators such as Windows `.\`) for `fsnotify` watch events
- This reduces cyclomatic complexity with no change in behaviour
- All existing tests pass

###Related issue(s)
Fixes #526 

###AI Disclosure
The code change was written and verified by me, I fully understand it.

### Testing
- `go build ./...` : passes


cmd/import.go
```diff
@@ -173,9 +173,7 @@ func NewImportCommand(globalClientOpts *connectors.ClientOptions) *cobra.Command
 					}
 
 					// Normalize file path to match the watcher fsnotify events format.
-					if strings.HasPrefix(f, "./") {
-						f = strings.TrimPrefix(f, "./")
-					}
+					f = strings.TrimPrefix(f, "./")
```
Vaishnav88sk: **Suggestion**: While dropping the redundant `if` statement is syntactically valid, `strings.TrimPrefix` is brittle for path normalization. It only removes a single `./` prefix and fails to handle redundant prefixes like `././dir/file` or platform-specific separators (e.g., `.\` on Windows). Instead of simple string manipulation, use the standard library's path utility to properly normalize the file path for `fsnotify` events:
```go
f = filepath.Clean(f)
```

Vaishnav88sk (dismissed):

cmd/import.go
ish-g09: Thanks for the suggestion, @Vaishnav88sk! That's a great point regarding Windows separators and redundant prefixes. I've updated the implementation to use filepath.Clean(f) and imported path/filepath.

Vaishnav88sk:
LGTM 👍🏻

Harsh4902 (approved):
GTG

PR #538, opened 2026-09-23
fix: honor active Docker context on macOS when /var/run/docker.sock is unavailable

### LINKED ISSUE #537: macOS: honor the active Docker context when /var/run/docker.sock is unavailable
### Describe the bug

`microcks start` cannot connect to Docker Desktop on macOS when Docker Desktop is configured without the privileged `/var/run/docker.sock` symlink.

Recent Docker Desktop installations can expose the daemon through the user-owned socket:

```text
unix://$HOME/.docker/run/docker.sock
```

The Docker CLI continues to work because it resolves this endpoint from the active desktop-linux context. However, microcks-cli creates its Docker client using client.FromEnv. When DOCKER_HOST is unset, the Docker SDK falls back to:
```text
unix:///var/run/docker.sock
```

This results in:
```text
failed to create container: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
```
Docker is running, and regular commands such as docker info and docker ps work correctly.

### Expected behavior

microcks-cli should connect to the endpoint configured by the active Docker context, matching the behavior of the Docker CLI.
An explicitly configured DOCKER_HOST should continue to take precedence.

### Actual behavior

microcks-cli tries unix:///var/run/docker.sock, even though the active desktop-linux context points to $HOME/.docker/run/docker.sock.

### How to Reproduce?

1. Install or configure Docker Desktop on macOS.

2. In Settings → Advanced, leave Allow the default Docker socket to be used disabled.

3. Ensure the Docker Desktop context is active:
```shell
$ docker context show
desktop-linux
```

4. Verify the context endpoint:
```shell
$ docker context inspect desktop-linux \
    --format '{{ .Endpoints.docker.Host }}'
unix:///Users/<user>/.docker/run/docker.sock
```

5. Confirm Docker works:
```shell
$ docker info
# succeeds
```

6. Ensure `DOCKER_HOST` is not explicitly set:
```shell
$ unset DOCKER_HOST
```

7. Start Microcks:
```shell
$ microcks start
failed to create container: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
```

_No response_

_No response_

### Additional information

_No response_

ish-g09:
### Description
- Added `ConfigureDockerHost` in `pkg/connectors/container_client.go` to help `microcks-cli` connect to Docker on macOS when the default `/var/run/docker.sock` isn't available.
- Checks the active Docker context first via `docker context inspect` so it honors whatever context the user is actively running (Docker Desktop, Colima, OrbStack, etc.).
- Adds a fallback to `$HOME/.docker/run/docker.sock` if context inspection isn't available.
- Leaves existing behavior completely untouched if `DOCKER_HOST` is already set or if `/var/run/docker.sock` exists (like on Linux).
- Hooked this check into `NewDockerClient()`, `PingDockerHost()`, and `cmd/testDryRun.go` so both `microcks start` and `microcks test --dry-run` can reach Docker smoothly on modern macOS setups.

### Related issue(s)
Fixes #537

### AI Disclosure
The code changes were written and verified by me. I fully understand them. Took help of AI to check for vulnerabilities. 

### Testing
- `go build ./...` : passes


ish-g09:
Hey @Harsh4902,

Just opened this PR for the macOS Docker context discovery, all CI checks are green.
Saw Copilot's summary mention unit tests. Happy to add a container_client_test.go covering DOCKER_HOST precedence and resolution if you'd like. Let me know what you think whenever you get a chance to look!

pkg/connectors/container_client.go
```diff
@@ -74,7 +75,44 @@ func NewContainerClient(driver string) (ContainerClient, error) {
 	}
 }
 
+// ConfigureDockerHost ensures DOCKER_HOST points to an active Docker daemon socket.
+// On macOS, Docker Desktop may not install the privileged /var/run/docker.sock symlink.
+// If DOCKER_HOST is not explicitly set, we look for the user socket or inspect the active context.
+func ConfigureDockerHost() error {
```
Harsh4902, line 81: Hey @ish-g09 This is a good workaround but if you see the codebase carefully, we don't set any `DOCKER_HOST` value for our docker client. We are using docker's official go-sdk to connect with docker. So ideally a solution should be from the docker sdk side and not a fallback on our side.

pkg/connectors/container_client.go
```diff
@@ -74,7 +75,44 @@ func NewContainerClient(driver string) (ContainerClient, error) {
...
+		userSocket := filepath.Join(homeDir, ".docker", "run", "docker.sock")
+		if _, err := os.Stat(userSocket); err == nil {
+			return os.Setenv("DOCKER_HOST", "unix://"+userSocket)
+		}
+	}
+
+	return nil
+}
 func NewDockerClient() (*containerClient, error) {
+	if err := ConfigureDockerHost(); err != nil {
```
Harsh4902, line 113: So, this check should be there only if docker's go-sdk dosen't provide the solution for this. Otherwise it a overhead on us to manage things whenever something is got changed on their end.

Harsh4902 (changes requested):

pkg/connectors/container_client.go
ish-g09: Hey @Harsh4902, totally fair point! I actually dug into the Docker Go SDK and re-read the Mac permission docs you linked to check if the SDK could handle this natively.
Interestingly, Docker's docs call out this exact scenario:


> "If you don't enable this option, the symlink... is not created and you may have to explicitly set the DOCKER_HOST environment variable to /Users/<user>/.docker/run/docker.sock in the clients it is using. The Docker CLI relies on the current context to retrieve the socket path..."


Basically, contexts are managed only by the docker CLI tool, while the Go client SDK (client.FromEnv) just looks for DOCKER_HOST and otherwise hardcodes /var/run/docker.sock. It doesn't know how to resolve contexts on its own.
That's why setting DOCKER_HOST through this small fallback (just like we already do for Podman in ConfigurePodmanHost) seemed like the simplest way to get it working on modern Macs without bringing in heavy CLI packages.

Let me know what you think! Happy to explore other ideas if there's a cleaner way you prefer or incase i've missed something.

Harsh4902: @ish-g09 I tried to dig deeper and found `Docker` has already deprecated it's original `github.com/docker/docker` go library which we are using currently. They are building a completely new [docker go-sdk](https://github.com/docker/go-sdk). And which provides access to low lever docker API and also have cli context so it can work smoothly. But, this sdk is still not in it's v1.0.0 which is basically first stable release. So the adoption of this library is crucial now. Let's think how long can we work with this temporary solution. Then we can adapt that. But for the long run, we anyways have to migrate to new docker go-sdk.

Harsh4902: I am also trying to connect with moby maintianers, as old docker go sdk is now transformed into [project moby](https://github.com/moby/moby). If they can provide some library level solution then it's well and good. Otherwise we have to go with temporary solution.

ish-g09: Makes total sense @Harsh4902! Happy to wait and see what they suggest. If it turns out they don't have a library-level option right now, we can always revisit this approach as a bridge until the new SDK stabilizes.

Keep me posted on what they say!

